PR #54, opened 2024-10-21
Better Sacrifice Teleporting

JohnnyDangerSpace:
## About the PR

Just changed the loop so it actually works like its meant to, and also will now refuse to send you into solars or off station.

## Why / Balance

It's just better teleporting so you don't get teleported into solars and die a horrible death.

## Technical details

Added it so it moves the player when testing to see if the spot works. This isn't visible through normal gameplay since its so fast so there's no reason to not do this. This allows me to check if the player has atmos and if they're on a grid.

## Requirements

**Changelog**

:cl:
- tweak: Getting sacrificed places you in better spots

### DIFF
--- a/Content.Server/_Goobstation/Heretic/Ritual/CustomBehavior.Sacrifice.cs
+++ b/Content.Server/_Goobstation/Heretic/Ritual/CustomBehavior.Sacrifice.cs
@@ -9,6 +9,7 @@ using Content.Shared.Damage;
 using Content.Shared.Heretic;
 using Content.Server.Heretic.EntitySystems;
 using Content.Server.Chat.Managers;
+using Content.Server.Atmos.EntitySystems;
 using Robust.Shared.Random;
 using Content.Shared.Mobs.Systems;
 using Content.Shared.Mobs;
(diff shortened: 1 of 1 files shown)


Content.Server/_Goobstation/Heretic/Ritual/CustomBehavior.Sacrifice.cs
```diff
@@ -213,11 +216,14 @@ private void TeleportRandomly(RitualData args, EntityUid uid) // start le telepo
             var randVector = randomSystem.NextVector2(maxrandomradius);
             newCoords = coords.Offset(randVector);
 
+            xformSystem.SetCoordinates(uid, newCoords); //move person teleported to check if they're under a tile (the tiles intersecting doesnt account for this so we need to do this) -space
```
taydeo, line 219: How did this look in your tests, was the person teleported around several times?

taydeo (changes requested):

Content.Server/_Goobstation/Heretic/Ritual/CustomBehavior.Sacrifice.cs
JohnnyDangerSpace: from my tests it works instantly and doesn't show them being teleported around the map

taydeo (approved):

PR #63, opened 2024-10-25
Reverts "In Memorium"

Icepicked:
Why are we "celebrating" this guy.

Also we dont need the Goob Server host immortalized on a downstream fork.

**Changelog**

🆑 
- remove:  Removed the reference to the man. We do not need to remember.


GitHubUser53123:
> In Memorium

> wtf why are we celebrating this guy

PR #69, opened 2024-10-30
Violet Alert added to Guidebook

DevilishMilk:
Because I forgot to add it :)

(built & tested it, it worked with no errors so if github denies this one i may cry)

:cl: add: Added Violet Alert to Alert Level SOP.


DevilishMilk:
no fuckin way

DevilishMilk:
did it fr stage the toolbox

PR #88, opened 2024-11-09
Slows down the Survival gamemode a bit

DevilishMilk:
## About the PR

Increased avg. round-end time in RampingStationEventScheduler (the survival scheduler) to 2 hours, which is about what we're aiming for with shifts. Current Survival modes end ramping up considerably after the 1 hour mark which usually ends up with a ninja, a few ratkings, and a dragon spawning within 10 minutes which, while often fun, is a little excessive for what should really be the midround.

**Changelog**

:cl: tweak: Slowed down Survival a little bit.


taydeo:
liberal wall of text

PR #96, opened 2024-11-14
Dynamic doesn't suck now - why was this not done months ago?

TooSillyFennec:
## About the PR
Dynamic is not laughably bad anymore, and _actually_ scales past the 20 player mark. 

## Why / Balance
Traitor objectives - 10 shits out 4-6 objective games regularly. 6 brings that back to the 2-4 we all know from back in the day and probably love. 

Midgame Changes - Shit actually depletes the threat level more than maybe 10 a game when it regularly is at 15-30, and major antags actually require the game to of dumped threat into mid game [Leaving very little for the start] or the shift to be at shitfest levels already.

Costs have generally been readjusted, with the solo antags generally being in the same realm without being EXACTLY the same- Lings aren't literally double traitors, just  a little extra, and heretics a little extra than lings- and on the other side thieves slightly less than traitors, not exactly the same.

Scaling has been "implemented" [An actual attempt requires a recode of dynamic and the entire gamerule system and 95% of the same effect can be done by .yml, not worth the effort really] by splitting the traitor gamerule into 3 [Low/Med/High] and Lings/Heretics into 2 [Low/High] - Low for scales as 2 max at 30 ratio, Traitor's medium and the rest's high scales as 4 max at 20 ratio [AKA an attempt at half the average game of lings/tots], and Traitor's High scales as 6 at a 15 ratio. Lings and Heretics break down at higher antag counts thus no larger variant.

The average game has around 40-60 Threat and 2/3rds often goes to game start antags- aka 30-40 threat, which has been set as the general balance point with tot medium/rest high set as around half or so of this level. Low's usually slightly below half and traitor's high slightly above 1.5x half. The intent's to make most games w/ 0 major antags 1/10 antags or so, mixing 2 or so rules with smaller and larger rules being rarer but making some variety. Thieves are excluded and stay as a low cost 2 max rule as they are a side antag, with the playerratio increased. Thieves weigh higher than ling/heretic but less than traitor, ending thief spam while keeping them common as they only spawn 2 max.

## Requirements

**Changelog**

:cl:
- fix: Dynamic Traitors have normal objective count now. Say "Byeeee!!!" to 5 kill objective games or the grand theft ripoff thief.
- fix: Dynamic actually scales past 20 players now.
- fix: Midround antags actually use more than... 3 threat... thanks, real smart. AKA, Midround threat level actually gets used and matters.
- tweak: Costs and weights for dynamic have been majorly reworked. Changelings don't eat crazy threat, Thieves don't spawn like crazy, and Traitors actually share codewords with more than one person and are the common antag they are supposed to be.

### DIFF
--- a/Resources/Locale/en-US/_Goobstation/game-ticking/dynamic.ftl
+++ b/Resources/Locale/en-US/_Goobstation/game-ticking/dynamic.ftl
@@ -5,12 +5,16 @@ dynamic-roundend-points-midround = Midround budget: {$points}.
 dynamic-roundend-gamerules-title = Executed Gamerules:
 
 dynamic-gamerule-unknown-title = Unknown Gamerule
-dynamic-gamerule-traitor-title = [color=crimson]Traitors[/color]
-dynamic-gamerule-changeling-title = [color=yellow]Changelings[/color]
-dynamic-gamerule-heretic-title = [color=purple]Heretics[/color]
+dynamic-gamerule-traitor-low-title = [color=crimson]Traitor - Low[/color]
+dynamic-gamerule-traitor-medium-title = [color=crimson]Traitor - Medium[/color]
+dynamic-gamerule-traitor-high-title = [color=crimson]Traitor - High[/color]
+dynamic-gamerule-changeling-low-title = [color=yellow]Changeling - Low[/color]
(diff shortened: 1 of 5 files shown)


taydeo:
this what im talkin about

PR #97, opened 2024-11-15
oops - SOP fix!

DevilishMilk:
Moved Cargo & Engineering SOP back down to SOP 101 :)

:cl:
- fix: SOP placement error in guidebook


taydeo:
liberal wall of text

PR #161, opened 2024-12-10
Vulpkanin tail wagging

RetroManb:
## About the PR
I was a bit upset that you can't wag your tail as a Vulpkanin at will when I realized that Einstein Engines has that, so I added it.

## Why / Balance
Good for role play, for example, to show when a character is happy without use of emotes or typing.

## Technical details
I renamed all vulpkanin tail wagging markings to have "Animated" at the end of them, and to have the same IDs as their non-wagging counterparts, so the WaggingComponent could work with it.
I know this breaks characters that were already using the wagging variants, I'm sure it'll be fine, but I'm also not sure how severely it would be broken.

## Media

https://github.com/user-attachments/assets/1e8b5bfe-187e-450f-861d-def90005f9d5

## Requirements

## Breaking changes

**Changelog**

:cl: 
- add: Vulpkanin tail wagging


RetroManb:
failed cuz "/Textures/_Goobstation/Changeling/bone_shard.rsi" has no state "festive", i never messed with goobstation's stuff so i think its fine

taydeo:
test fail unrelated

PR #186, opened 2024-12-19
Secret antag rate changes

JohnnyDangerSpace:
## About the PR

Added blob as a normal spawning gamemode in secret, and added blobtots (traitors and blob) as well. I also removed the heretic gamemodes and reverted them back into a side antag like thief.

## Why / Balance

First off, heretic needs to be a side antag on the station. While they can be really powerful, they suck extremely bad without someone or something else getting secs time and attention. As well, they really don't work together as well, so having 4 of them really doesn't do much but make crossfire in the objectives. Don't get me wrong, I get the idea of having them in their own gamemode, but i think it would work better here as a side antag in any mode with a max of like 1 or 2 depending on how many tay wants. (tldr, heretics need to sacrifice like 4 people, tots at most have to kill like 1 on average, its really stacked against them when its all counterplay on them)

Second off, blob is awesome and i have no idea why it isnt in the secret gamemode. I also added totblob just for a bit of extra fun. I was considering bloblings but lings would be too powerful against blob via their regen.

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added blob to secret!
- tweak: Changed heretics to spawn as a side antag like thief.

### DIFF
--- a/Resources/Prototypes/GameRules/roundstart.yml
+++ b/Resources/Prototypes/GameRules/roundstart.yml
@@ -22,8 +22,8 @@
     rules:
     - id: Thief
       prob: 0.75
-    #- id: Heretic # goob edit
-    #  prob: 0.4 # funkystation - nuh uh
+    - id: Heretic # goob edit
+      prob: 0.4 # funkystation - readd side antag heretic
 
 - type: entity
   id: DeathMatch31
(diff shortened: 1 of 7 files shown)


JohnnyDangerSpace:
also tay make sure you double check i readded heretics as a side antag correctly, I didn't have time to check your pr where you removed them.

taydeo:
CI test is related.

PR #203, opened 2024-12-23
Hotsprings

hereelabs:
## About the PR
This PR contains the "Hot Springs" event map and the Party Bus shuttle which will be used by admins to transport crew members to said map. The map itself is a planet-side map containing a normal air mix atmosphere. Essentially, it is a RP map taking place planet-side. The premise of this event is that the Syndicate / Cybersun is sponsoring it under the guise of a vacation, to which Central Command somehow agreed (comically so). The crew is sent there after 5 minutes of normal gameplay, where they will spend the remainder of the shift playing on. The map contains plenty of activities, and the entire Service department is set up, meaning Botany, Kitchen, and the Bar all are usable. This is my first PR on Github, so I imagine there may be some mistakes.

## Why / Balance
The Hot Springs map, I believe, gives a great role play opportunity, even if it's ran as a one time event. Planet-side maps are very uncommon, even as events, so I believe this would give a breath of fresh air from the constant 3 hour shifts of attempting to fight back whatever Secret throws at us.

## Technical details
Creates a folder called `Event` in /Resources/Maps/
Adds `hotspring.yml` and `partybus.yml` to the /Resources/Maps/Event/ folder.
Adds a map prototype for the Hot Springs map, which only exists in the case it is needed for some reason. It is not meant to be a normal station map, but if forcemapped should (partially) work.

## Media
SS14 Loader_LPCLkZ1y1V

SS14 Loader_4A4PSGhkOi

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Planet-Side Hot Springs map and Party Bus Shuttle for "Hot Springs" Event


taydeo:
image
Add your map to the test cases.

hereelabs:
added the map to the test cases

PR #220, opened 2024-12-25
Flash & Baton buff

DevilishMilk:
Stun batons are now a 2 hit stamcrit if timed well, and flashes now stun for double duration (1.5s -> 3s), allowing a flash cuff if timed quickly enough.

Why?
Sec equipment kinda sucks ATM. Buffing Sec equipment can lead to more fun antag equipment down the road so everyone has better tools to deal with everything without admin intervention.

:cl:
- tweak: Stun batons now stamina-crit in two hits rather than three
- tweak: Flash stun duration doubled to 3 seconds


kbarkevich:
I McLike this

PR #243, opened 2024-12-30
The Prayerening

DevilishMilk:
Copied fax code so that prayers also give admins an alert so we don't miss prayers or rings (abusing this feature may result in a smite)

Not adding it to the changelog because I don't want casuals to know


Erisiaa:
BALKEN RAGE + STILL WATER + NOADRENALINE + NO AURA + THOSE WHO KNOW 💀 💀 💀 💀

PR #309, opened 2025-01-21
Players can no longer crawl over tables

DevilishMilk:
RAHHHHHHHHHH Title

Also, plastic flaps are now MidImpassable (medium objects can pass through fine)

:cl: Miiish
- fix: Players can no longer crawl over tables


supersayangoblin:
I don't really get the complaint over going under flaps... If people are using crawling to get to restricted areas then mappers should add windoors past the flaps, it's not like you can't just unwrench it anyways.

DevilishMilk:
For the record, this change only applies to tables - flaps can still be passed under just fine.

PR #315, opened 2025-01-23
Nerfed the Medical Multitool's doAfter Speed

kbarkevich:
## About the PR
The Medical Multitool's doAfter speed (the speed at which the loading bar progresses) has been changed from 2x the normal tool speed to 0.75x the normal tool speed.

Credit to Cthyga on the Funky Station Discord for the original idea for this nerf.

## Why / Balance
If the normal tools are considered the baseline of speed, then there are only two ways they could be improved. One could either:
1) Increase the speed of the loading bar (shorter doAfter speeds), shortening the amount of time it takes to complete a single step
or
2) Combine multiple surgical tools into one object, decreasing the time it takes for you to move from one step to the next.

Going beyond the basic surgery tools, a set of upgraded tools already exists that accomplishes the first possibility (the Advanced Retractor, the Energy Scalpel, the Searing Tool, etc). These tools are specialized, but perform their surgeries much faster than their basic counterparts.

The Surgical Multitool is an object that should accomplish the second possibility. By combining every surgical tool into one, you will save _enormous_ amounts of time that would otherwise be spent juggling objects around, closing and re-opening menus, etc. The Surgical Multitool eliminates all that and makes every surgery a matter of clicking one button after the next.

It does not make sense that the Surgical Multitool, an item that accomplishes the second possibility extremely well and already saves tons of time on that point alone, should also not only have shorter doAfter times than the normal surgical tools, but also **shorter doAfter times than the highly specialized, UPGRADED surgical tools.**

If the upgraded surgical tools exist for the purpose of allowing a surgeon to perform those specific tasks at very high speeds, then the Surgical Multitool should not be allowed to make them redundant. This nerf ensures that the Surgical Multitool remains highly valuable as a one-size-fits-all problem solver that still simplifies all surgeries down to clicking a series of buttons, while simultaneously ensuring that the more specialized, upgraded surgical tools still have a reason to exist.

## Technical details
Changed some numbers.

## Media
N/A

## Requirements

## Breaking changes
N/A

**Changelog**

:cl:
- tweak: Reduced Surgical Multitool doAfter time.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Specific/Medical/surgery.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/Medical/surgery.yml
@@ -445,25 +445,25 @@
     startSound:
       path: /Audio/_Shitmed/Medical/Surgery/saw.ogg
   - type: Hemostat
-    speed: 0.75
+    speed: 0.75  # funkystation
   - type: Scalpel
-    speed: 0.75
+    speed: 0.75  # funkystation
   - type: Drill
-    speed: 0.75
+    speed: 0.75  # funkystation
   - type: BoneSetter
-    speed: 0.75
+    speed: 0.75  # funkystation
   - type: Retractor
-    speed: 0.75
+    speed: 0.75  # funkystation
   - type: Cautery
-    speed: 0.75
+    speed: 0.75  # funkystation
   - type: BoneGel
-    speed: 0.75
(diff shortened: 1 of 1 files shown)


taydeo:
Mark all your changes with `# funkystation`

taydeo:
lol ok

PR #316, opened 2025-01-24
Removes Antag scheduler from Blob gamemode

DevilishMilk:
## Why

Now in-line with the rest of the gamemodes so events aren't firing double what they should be
## Technical details

Removed a line


taydeo:
INSTANT MERGE

PR #324, opened 2025-01-24
Revs: Evac shuttle doesn't arrive when HRevs are alive

taydeo:
# This PR is part of a larger Revolutionaries rework.
Any PR with this as the header is considered a part of the larger revs rework at Funky Station.

## About the PR
When Command calls the evac shuttle, after the wait period ends, the evac shuttle doesn't arrive. Instead, a message is broadcast that the evac shuttle could not dock because of strange activity on the station. Also with this change, when all of command is dead, the round just ends. (We understand that rounds with no command will just auto-end, too bad too sad.)

This PR also changes `revolutionaries-win-sender` to Yucatan Communications, to be more in-line with Funky lore.

## Why / Balance
Too many half measures and half baked features make revs terrible to play on Goob down-streams. Having the communications console just not be able to call evac makes the game-mode so easy to meta. The terror and panic (or wonder and excitement!) that ensues when you realize that your station is full of revolutionaries is what makes this game-mode good. This should also allow rounds to end nicer and more gracefully, as there is a 2 minute cool-down to the round ending when revs win.

Also, revs going to centcomm is stupid asf idc what anyone says.

We also hope that this encourages HRevs to play things safer, and to stop charging into security with no armor expecting a different outcome other than being gunned down immediately.

## Technical details
Adds a new event to `EmergencyShuttleSystem` called  `ShuttleDockAttemptEvent`, which quite frankly was missing on upstream. Should be obvious when this is called in-game (when the evac shuttle... attempts to dock lol)

## Breaking changes

idk lol

**Changelog**

:cl: taydeo, Rainbeon
- tweak: The Evacuation Shuttle no longer Docks when HRevs are still alive
- tweak: The game now auto-ends when revolutionaries win.
- tweak: The Evacuation shuttle is now auto-called when HRevs die.

### DIFF
--- a/Content.Server/GameTicking/Rules/RevolutionaryRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/RevolutionaryRuleSystem.cs
@@ -31,10 +31,12 @@ using Content.Shared.Cuffs.Components;
 using Content.Shared.Revolutionary;
 using Content.Server.Communications;
 using System.Linq;
+using System.Threading;
 using Content.Shared.Chat;
 using Content.Server.Chat.Systems;
 using Content.Shared.Changeling;
 using Content.Shared.Heretic;
+using Timer = Robust.Shared.Timing.Timer;
 
 namespace Content.Server.GameTicking.Rules;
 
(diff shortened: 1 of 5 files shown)


Content.Server/GameTicking/Rules/RevolutionaryRuleSystem.cs
```diff
@@ -31,10 +31,12 @@
 using Content.Shared.Revolutionary;
 using Content.Server.Communications;
 using System.Linq;
+using System.Threading;
 using Content.Shared.Chat;
 using Content.Server.Chat.Systems;
 using Content.Shared.Changeling;
 using Content.Shared.Heretic;
+using Timer = Robust.Shared.Timing.Timer;
```
Rainbeon, line 39: Didn't we end up not making use of this?

PR #363, opened 2025-02-07
Rebalanced weights of antags & subgamemodes

DevilishMilk:
## About the PR
Decreased the rate of survival & zombie modes, made blobtraitor more common than blob (we typically only get one blob and it RARELY gets anywhere)
Survival and zombie modes are generally just shitshows and we've been getting enough pop for midround zombies to become common enough that I wouldn't even be particularly worried about removing the gamemode entirely (not needed yet)
Re-added traitor revs as a treat
Set heretics to spawn as 20% prob instead of every other round (I still think it's too high but eh)
Fixed heretic rolling under 8 pop setting the lobby to extended (it was 1 bool change)
Set thief ratio to 20 players instead of 25 as a treat
I want to see how these go but I think if these changes do fine we can consider decreasing the player ratio for most antags, mostly traitors

:cl: Miiish
- tweak: Lowered rates of zombie & survival modes
- fix: Fixed the heretic subgamemode breaking the lobby at lowpop


DevilishMilk:
Another day another garbage PR

taydeo:
Mald PR

DevilishMilk:
Another day another garbage branch update

PR #390, opened 2025-02-15
Round End Stats

empty0set:
~~Stolen~~ _borrowed_ from upstream PR https://github.com/space-wizards/space-station-14/pull/30686

This is something that has been brought up (in concept) multiple times.  Opening this as a draft to see if this is a good start.

## TODO:
- [ ] Actually add things to keep track of
  - And is it even worth it to do this downstream?  Touch a bunch of files -> create merge conflict hell
- [ ] Think of a way to .yml-ize specifying stats?
- [ ] Some way of choosing particularly interesting stats out of a list
  - Since right now it will just print out all the stats unconditionally.  Conceptually it should be outliers like "Clown was slipped 3423 times" (oh my god what the hell happened this round?) instead of "Clown was slipped 2 times" (oh my god who cares?).  And it should be limited to a couple items so people will actually read them.
- [ ] Some way of having flavor-text for certain ranges--for that added ambiguity of what's being tracked
  - Like if the station budget is less than 2000 or something: "You're all fired!", or if the station budget is some gigantic number: "Congratulations crew!  NT will be rewarding your hard work with a pizza party next week!", etc.


lzk228:
> And is it even worth it to do this downstream?

nuh

empty0set:
> > And is it even worth it to do this downstream?
> 
> nuh



> > And is it even worth it to do this downstream?
> 
> nuh

Glad to see that this is still being worked on! I hope it makes it in upstream as it's a really good idea.

PR #429, opened 2025-02-22
Funky NightVision Item implementation

JoulesBerg:
## About the PR
Added onEquip, onUnequip, and onShutdown event handlers to NightVisionComponents
In order to not break / over complicate current implementation uses of the component type I just made it so the component gets added and removed from the wearer following the appropriate event, along with the action as necessary.

## Why / Balance
people were asking for it, personally I don't care either way and I've invested enough time trying to figure out the cursed nightvision implementation and don't feel a desire to invest more time at this point. but its here for anyone who wants to build on it or get it to 100% functionality instead of the 90% functionality its at right now.

## Technical details
added event subscriptions to shared nightvision system and moved the nightvisionchangedevent to the shared component file so it can be imported to the client if needed. 

## Media
No in-game changes, just added functionality for anyone who WANTS to add NightVision to certain items. I won't begin to pretend I understand the balance implications of item fuckery.

## Requirements

## Breaking changes
It should be noted that just due to the way adding components works, and the overcomplicated spaghetti that is EntityUid's and Components - plus a lack of simple 1 - 1 functionality between the two types -  that if something that already has nightvision puts on an item granting night vision, then removes the item, their original night vision will likely ALSO be removed. This can probably be fixed with a simple boolean inside the component and adding a check to the unequipped event, maybe with a loop that goes through all the NV components in the wearer. 
It should also be noted that due to something I wasn't willing to invest the time to figure out the reasoning behind, the CLIENT component will always use the default color green for item nightvision. 

**Changelog**
clothing items like sunglasses can now use `- type:  NightVision` argument in their yaml to give toggleable or non toggleable nightvision to the wearer.

### DIFF
--- a/Content.Client/_Goobstation/Nigthvision/NightVisionSystem.cs
+++ b/Content.Client/_Goobstation/Nigthvision/NightVisionSystem.cs
@@ -20,6 +20,8 @@ public sealed class NightVisionSystem : EquipmentHudSystem<NightVisionComponent>
     {
         base.Initialize();
 
+        SubscribeLocalEvent<NightVisionComponent, AfterAutoHandleStateEvent>(SyncClientComponent);
+
         _overlay = new(Color.Green);
     }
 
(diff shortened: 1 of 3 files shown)


Frederick-Simon:
Hey thank you so much for this, I'll try and implement it at some point soon, looking at Ninja goggles mainly

McBosserson:
I suggest closing this PR, since we got a form of night vision.

PR #454, opened 2025-02-27
Made crew definition in crewsimov explicit as part of lawset, changed emag "only X are Y" to be more universal.

misterghast:
## About the PR

Crewsimov has had an extra law added that defines crew as "those who are visibly carrying or wearing a NanoTrasen-Brand ID card."

Emag "only x are y" law has been changed to "X and people they designate as such are Y and only they are Y.", to make it more universally applicable and work with the new crewsimov law, hopefully without confusion.

## Why / Balance

It's weird to me for a specific lawset to have predefined definitions but not others. Crew definitions could be justified in-character as preprogrammed into borgs, but it's just kind of annoying to me that this is the only lawset that this is done with. I understand it's because it's default but if this doesn't harm anything, it seems worth doing. It could also open up reinterpretations of the definition of crew by making it a law instead of a rule - And therefore able to be reinterpreted without it just being shitter behavior, as long as the interpretation makes sense. I'm not sure what other interpretations could be done here though, it seems explicit. 

## Technical details

It was mostly just the addition of one law, shifting of the rest in crewsimov down, and then entering the new stuff in language files and shifting it down there.

Emag law was just one change in the language files.
Translations not added yet sadly.
## Media

image

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Added extra law to Crewsimov lawset that defines crew.
- tweak: Changed emag lawset to be more widely applicable and less vague with more lawsets.
-->


mkanke-real:
No self preservation law seems quite odd to me aka the old law 3 "You must protect your own existence as long as such does not conflict with the First or Second Law." Was there a reason you removed that one instead of just making it 4 laws standard with the one you are proposing in say the 2 slot? I mean I do see a scroll bar is the self preservation hiding down further?

misterghast:
> No self preservation law seems quite odd to me aka the old law 3 "You must protect your own existence as long as such does not conflict with the First or Second Law." Was there a reason you removed that one instead of just making it 4 laws standard with the one you are proposing in say the 2 slot? I mean I do see a scroll bar is the self preservation hiding down further?

Yes, the self preservation law is the new law 4 and is just hidden in the extra scrollable space in the screenshot.

taydeo:
Requires further review by other maintainers.

corresp0nd:
not a fan of the emag law, it's really just repeating the same thing twice. i would rather the laws be concise and easy to understand at a glance, rather than them covering every possible avenue. 

the new crewsimov law is. Eh. i do think there is a minor issue with how crew is defined because it usually requires some sort of OOC clarification, which is generally bad for new players. but also, i think "rules lawyering" is one of the bigger draws of borgs. you should be arguing, in character, about your laws and how you carry them out. i _want_ syndies to lie to borgs about who's who, and i want borgs to be able to roll with that. 
also, i feel like this sort of concrete idea of "who is crew" ruins a lot of IC interactions. there have been problems with borgs crowbaring crewmates who lost their ids to death, this would make that worse. i also feel like "very visible antag who happens to have a crew ID" would become a major problem (lings in general, agent ids, etc). 

there is no perfect way for a borg to follow their laws, the way they carry out their laws can vary per round as long as they can somehow argue that it fits. i feel like this solution causes more harm to this than it fixes anything

taydeo:
Thank you for your contribution, but this PR will not be merged per the explanation above.

PR #477, opened 2025-03-03
Reverts the 'Reversion' done to the rebalance to disablers

kbarkevich:
This 'Reversion' is just a mald pr from someone upset at people using non-lethals. --Reverts commit fe2f7f7

## About the PR
Reverts the changes made in commit fe2f7f7 on funky station.

## Why / Balance
Undoes a downstream "Rebalance" that in others and my opinions just makes disablers and the disabler SMG underpowered.

In all seriousness -- a weak disabler makes Security toothless to combat anybody for whom lethals are not authorized. The reversion this PR undoes would serve to make Security generally unable to effectively fight people without either being

1) Very robust (which sec isn't supposed to have to be, thus why they are allowed to carry this sort of equipment)

or

2) Willing to use lethals against people they should be disabling (which nobody should want)

Discouraging Security from using non-lethal modes of combat is a bad idea. Gatekeeping baseline-levels of Security duty to people who are really good at combat in this game is a bad idea. Reducing Security's ability to non-lethally fight more than one opponent at a time is a bad idea. Making Security weaker while we are already experiencing shortages of players willing to roll for Security roles is a bad idea. Making digs at people contributing to the game by saying they're just being "mald" is a bad idea.

## Technical details
just changed the values in the two yml files to what they were before the reversion merge

## Media
pew pew

## Requirements

- [ ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
N/A

**Changelog**

:cl:
- tweak: Re-strengthened disablers

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -453,7 +453,7 @@
       path: /Audio/Weapons/Guns/Gunshots/taser2.ogg
   - type: ProjectileBatteryAmmoProvider
     proto: BulletDisablerPractice
-    fireCost: 100
+    fireCost: 50
   - type: Tag
     tags:
     - Taser
@@ -481,7 +481,7 @@
       - Belt
   - type: ProjectileBatteryAmmoProvider
     proto: BulletDisabler
-    fireCost: 100
+    fireCost: 50
   - type: GuideHelp
     guides:
     - Security
@@ -508,7 +508,7 @@
         shader: unshaded
   - type: Gun
     selectedMode: FullAuto
-    fireRate: 4
+    fireRate: 4.5
     availableModes:
       - SemiAuto
       - FullAuto
@@ -516,7 +516,7 @@
       path: /Audio/Weapons/Guns/Gunshots/taser2.ogg
   - type: ProjectileBatteryAmmoProvider
     proto: BulletDisablerSmg
-    fireCost: 33
+    fireCost: 25
   - type: MagazineVisuals
     magState: mag
     steps: 5
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
@@ -228,7 +228,7 @@
       fly-by: *flybyfixture
   - type: Ammo
   - type: StaminaDamageOnCollide
-    damage: 30
+    damage: 33
   - type: Projectile
     impactEffect: BulletImpactEffectDisabler
     damage:


Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
```diff
@@ -508,15 +508,15 @@
         shader: unshaded
   - type: Gun
     selectedMode: FullAuto
-    fireRate: 4
+    fireRate: 4.5
```
taydeo, line 511: whats this about

kbarkevich: It undoes a change the original PR this responds to made. Check that PR, it changed it from 4.5 to 4.

taydeo: :skull:

taydeo:
I do generally agree with the ideas here. The original PR was merged because of me reverting this PR in the original upstream merge I attempted to do to allow maintainers and other experienced contributors to comment on it, but that's whatever.

Generally, we want to buff securities tools to allow more room for wackier and more insane things to do with antags. Security isn't a department who's function is to serve as training targets for all of our super cool antags or whatever, they have an actual importance to how a round flows and how a story is told.

Security was woefully under-powered for a long time, and yes, the original reversion PR does help in making them under-powered. Do I agree with the exact numbers here? Dunno, we'll have to see, but one this is for certain though, reducing securities chances in detaining someone peacefully goes against what we're trying to do here.

taydeo:
anyway speed merge when

JoulesBerg:
My thoughts, even though its already merged:

- Fire rate is how many times per second the weapon can fire IIRC, though thats the number I know the least about. In this specific change I think only the smg fire rate is changed.
- I resent the (possibly sarcastic) implication that my PR was a mald PR, since I was just doing something there was actually a legitimate community discussiom about and consensus that the buff was bad. Calling the original commit a mald pr was probably in bad taste, but that was the title of the original discussion on discord. Regardless thats neither here nor there, tay has final say and if the desire is to buff sec then i don't have a problem against the idea in general.
- that being said, heres why the community (myself included) think the buff is bad.
Specifics of the buff:
1. Pre buff it took 4 shots to disable someone? Thats not that difficult, but the damage buff from 30->33 (damage is applied to stamina) makes that 3 shots instead. Thats a 25% buff (1.25)
2. Pre buff the base disabler had 10 shots. Now it has 20. 100% buff (2.5x total buff to the base disabler) 
3. Pre buff the disabler smg cost 33 per shot, now its 25 which is a 32% buff (1.65x buff so far)
4. Pre buff the disabler smg only shot 4/s now its 4.5, 12.5% buff (bringing the total smg buff to ~1.85x)
The consensus in the discussion on discord was that disabler was ALREADY overpowered before the buff, and you can carry a second disabler if you really need the extra ammo.

I don't know enough of the nuances to tell you everything about why the disabler is OP, but I felt in general when playing sec that it was a very effective weapon pre buff (and I am not a sec main). I don't think there are resistances to the disabler on most armor types, theres a video on liltenheads channel of a replay on harmony where death squad gets called and later on in the fighting the CE uses a disabler at close range to 1v1 and quickly stun a death squad member who got seperated and steal their pulse rifle. I think this was possible due to the lack of resistances, but again I don't fully understand the nuances, just that most everyone on the discord who commented agreed that this buff to the disablers was BUSTED.

If the intent is to buff sec I don't think this is the right way to do it. The disabler pre buff was already very effective at its base job, which was to stun the thief or other non violent antags and bring them in for questioning. This buff is extreme overkill in my, and other community members opinions, which is why I made the original reversion PR.

To address the concern of buffing sec/giving them better non lethal options, idk. But personally I think this huge buff to disablers, while it does do wonders for making officers want to use non lethals, also potentially makes the disabler and disabler SMG a better option in certain situations where sec SHOULD NOT be using non lethals. 

Especially the SMG, which, against any civilian that you should be using non lethals on, is an absolute beast already, without the almost 2x buff.

There should be a learning curve to sec, idk how steep it should be, but it shouldn't be easy. Maybe the disablers do need a buff, or there needs to be another option like tranq darts or something. 

Personally I think the biggest issue I have seen with sec is that interns often get treated like normal sec officers. I only ever had someone in sec offer to teach me as a sec intern once, and by that point I was already 3-4 rounds into my sec journey and had needed to figure most stuff out myself. 
Idk if this is just because interns look almost exactly the same as normal sec officers, or what, but yea. Sec should be taking noobies to the firing range, giving them practice disablers, giving them tutorials on their stun buton, telling them in looc how to do a wide swing, etc. 

Don't solve the problem by making an already busted weapon in the hands of someone who has an idea what they are doing even more busted. Solve it by making sec role play more so their interns actually get taught how to play the game and do combat. Or solve it with a less absolutely over the top buff.

PR #488, opened 2025-03-06
Heretic multi-patch and dejankification 

misterghast:
## About the PR

The focus glitch has been fixed, as the heretic magic ability check not uses the inventory RelayEvent instead of the HandSystem RelayEvent. I have also made it so that the Mansus Grasp can be toggled with the ability (turning it off does not activate it's cooldown, unlike turning it on), and that the magic contraband text of heretic items now properly displays instead of the localization string. 

## Why / Balance

The mansus grasp tweak was basically already in the game (you can just use it on any old non-living object to get rid of it) and in the few situations where you couldn't put it away, you really should have been able to.  The rest is clearly a restoration of intended behavior. 

## Technical details

Made mansus grasp action delete the grasp when used with it active. This does not set it as handling the action, so the action does not get put on cooldown from this.

The contraband issue seemed to have been caused by a purposely null line in the resource/definition/whatever file, which was probably a holdover from that line needing to be included.

The focus glitch was fixed by just making it check the player's inventory as well as the hand, by adding the SharedInventorySystem.Relay to the subscribers for the magic focus check event. 

## Media

image

## Requirements

- [ x ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ x ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

There is a possibility that the contraband text fix could conflict with what was done for the blood cultist contraband, but that should be easy to resolve.

**Changelog**


Content.Server/_Goobstation/Heretic/EntitySystems/HereticMagicItemSystem.cs
```diff
@@ -24,11 +24,11 @@ private void OnCheckMagicItem(Entity<HereticMagicItemComponent> ent, ref HeldRel
     private void OnCheckMagicItem(Entity<HereticMagicItemComponent> ent, ref InventoryRelayedEvent<CheckMagicItemEvent> args)
         => args.Args.Handled = true;
 
-    private void OnMagicItemExamine(Entity<HereticMagicItemComponent> ent, ref ExaminedEvent args)
+    /*private void OnMagicItemExamine(Entity<HereticMagicItemComponent> ent, ref ExaminedEvent args)
     {
```
taydeo, line 28: Do not comment code. Just delete it.

Content.Server/_Goobstation/Heretic/Abilities/HereticAbilitySystem.cs
```diff
@@ -145,7 +146,15 @@ private void OnMansusGrasp(Entity<HereticComponent> ent, ref EventHereticMansusG
 
         if (ent.Comp.MansusGraspActive)
         {
-            _popup.PopupEntity(Loc.GetString("heretic-ability-fail"), ent, ent);
+            //_popup.PopupEntity(Loc.GetString("heretic-ability-fail"), ent, ent);
```
taydeo, line 149: Same here. Do not comment code, just delete it.

taydeo (changes requested):

JoulesBerg:
I would also say leaving the action unhandled on cancel could cause problems. I haven't looked specifically at the mansus grasp code to know if it WOULD, but I think its a bad practice in general. Leaving events or actions unhandled I think should be avoided - some events and actions run in a continuous loop until handled, which could use up resources, etc. In some cases this could even be a memory leak (situtation where the client doesn't realize theres an unhandled event, or there is no longer a legitimate code path to mark the event as handled)

Personally I would say its better to just mark it as handled and initiate the cooldown - this would also act as a deterrent / punishment for just randomly activating the ability.

misterghast:
OK. I removed the commented code completely and changed how the manus grasp's cooldown works so the majority of the cooldown is now applied when the actual grasp is used (there is still a two second cooldown for enabling and disabling it, to stop spam). I did this instead of just applying the full cooldown on disable because that would encourage users to mansus grasp inanimate objects which I want to avoid.

taydeo (approved):

PR #529, opened 2025-03-17
Got rid of random crate

cubbyduck:
Removed the random crate in maildrobe on saltern
Make sure you check this because I'm very new to mapping


cubbyduck:
To specify it was a crate that was intersecting with the maildrobe

PR #568, opened 2025-03-23
Atmos Borg Module

SpaceCat-Chan:
## About the PR
added an Atmos borg module which is researched under the advanced tools industry research

## Why / Balance
Borgs are significantly less able to do Atmos than players are, due to not having an RPD and being unable to use the holofan, resulting in numerous unintentional spacings when they attempt to do atmos

## Technical details
todo?

## Media
todo.

## Requirements

- [ ] I have added media to this PR or it does not require an ingame showcase.

## Todo
- [x] get actual sprites instead of hue-shifted RPD module
- [x] figure out whether is actually should be researched like this, unlocked from the start, or be a default part of the engineering borg
- [x] make sure i actually did this correctly

**Changelog**

:cl:
- add: Added an Atmos Borg module containing an RPD and Holofan Projector, which can be researched by science and then constructed in the exosuit fabricator.


SpaceCat-Chan:
The holofan projector in the module has no way of being recharged, and making a holofan with infinite charge and where you can't take out the battery (there is no way to put the battery back in) seems like it's going to require either changing the holoprojector system or making another, someone please provide guidance

Resources/Prototypes/borg_types.yml
```diff
@@ -48,6 +48,7 @@
   - BorgModuleConstruction
   - BorgModuleRCD
   - BorgModuleCable
+  # add atmos module here?
```
taydeo, line 51: Atmos module should be researchable by sci, so it should not be a default.

PR #582, opened 2025-03-27
disable changeling transform sting

DevilishMilk:
## About the PR
Removed the transform sting from the changeling catalog

## Why / Balance

It's generally unenjoyable to get transformed at the start of the round and be stuck like that all round.

Ideally, this would be temporary or there would be a way to revert this, through Genetics or Chemistry. For the time being, it doesn't really fit Funky.

**Changelog**
:cl: Miiish
- remove: Removed transform sting from the changeling catalog


DevilishMilk:
Mald PR

taydeo:
Mald speed merge

JohnnyDangerSpace:
stale station has fallen

PR #614, opened 2025-04-06
Add remaining /tg/ gases

marc-pelletier:
## About the PR
This PR adds the remaining /tg/ gases and their reactions 

Hyper-Noblium:
Extremely inert, Hyper-Noblium stops other gases from reacting. (Specifically, it stops reactions when >5 moles). The rarest and most precious gas in the game and for good reason. Can be created when Nitrogen is combined with Tritium at extremely low temperatures. BZ works as a catalyst. 20:10 Nitrogen and Tritium respectively is consumed when one mole of Hyper-Noblium is condensed. Counter to the Nitrium reaction, BZ is not a requirement at all. But rather, it decreases the energy released by your gas mix during the reaction.

ProtoNitrate:
Proto-Nitrate is a highly reactive gas, but non-toxic when breathed. It is created in an exothermic reaction when Pluoxium is exposed to H2 at temperatures between 5000-10000 K. Hydrogen is consumed at around 10x the rate of Pluoxium.

When between temperatures of 260-280k, Proto-nitrate reacts with BZ to produce Nitrogen, Helium, and Plasma in an exothermic reaction. When between temperatures of 150-340k, Proto-nitrate reacts with tritium to produce H2 and radiation in an exothermic reaction. Proto-Nitrate reacts with at least 150 moles of H2 to create more Proto-Nitrate in an endothermic reaction.

Zauker:
Zauker is an incredibly deadly gas if inhaled. It is made by mixing Hyper-Noblium and Nitrium in an endothermic reaction at temperatures between 50000-75000 K. Nitrium is consumed at around 50x the rate of Hyper-Noblium. It is worthy to note that Hyper-Noblium stops reactions when it is present in quantities above 5 moles, prepare accordingly!

Zauker also decomposes exothermically into a 30/70 O2/N2 mix when exposed to Nitrogen.

Halon:
Halon acts as a fire suppressant by removing oxygen in the air (while producing CO2) in an endothermic reaction if the air temperature is above 100 C or 373.15 K. The oxygen suppression rate is 20 O2 : 1 Halon. It is created in a slightly exothermic reaction between CO2 and N2O in turfs with an active electrolyzer on them, below 230K, and at low pressure. 2 moles of CO2 are used and 1 mol of N2O is used.

Helium:
Helium is an invisible, inert gas. Functionally useless. 

Anti-Noblium:
Anti-Noblium is a rare gas used for not very much at all. It does look pretty when in the air though! Must be why cargo gets so much for it. 

## Why / Balance
These gases, aside from perhaps Hyper-Noblium, won't really have much of an impact on typical atmos rounds, which is why I am including them all in a single PR. These gases will help pave the way for including more /tg/ content in the future, but on their own, are not particularly impactful. 

## Technical details
All reactions follow original /tg/ formulas as closely as possible. 

## Media
image

## Requirements

## Breaking changes
None.

**Changelog**
:cl:
- add: Added Hyper-Noblium, an extremely inert gas that can stop other gases from reacting.
- add: Added ProtoNitrate, a highly reactive, but non-toxic gas that reacts with many other gases.
- add: Added Zauker, an incredibly deadly gas if inhaled, though dissipates quickly when mixed with nitrogen.
- add: Added Halon, a gas that acts as a fire suppressant by removing oxygen in the air while producing CO2.
- add: Added Helium, an invisible, inert gas. Can be sold to cargo for pocket change.
- add: Added Anti-Noblium, a rare gas used for not very much at all, but can be sold through cargo for a large sum of money.


AverageMoron:
how are you expected to hit 24K for hypernob? in 13 it was done using tier 3 parts but we dont have a system like that here outside of hellfire freezers

marc-pelletier:
> how are you expected to hit 24K for hypernob? in 13 it was done using tier 3 parts but we dont have a system like that here outside of hellfire freezers

In 13, hypernob was made at temperatures starting at 15K. I increased it to 24K to give people a slim chance of being able to make it using frezon cooling and hellfire freezers. If we ever get advanced parts, it will be going back down to 15K.

Terkala:
1: I heard that the gases were a big source of lag, so that's my first thought with adding more.
2: Will make setting up scrubbers take... a long time now. Lots of gases to click on-off
3: Hyper-Nobilum makes almost all frezon machines that I know people currently use stop working. Most of the time when you make frezon, the fact that you have N2 in the mix automatically makes the system drop to 23kelvin from making N2O.
4: ProtoNitrate will never get made, since it's impossible to handle and only is useful to kill people. A combination of making heat and rads means only a CE can safely handle it (all suits that protect from rads don't protect from heat, and vice versa).
5: Zauker is unlikely to ever get made, since it requires hyper noblium and nitrium, two hard to manufacture gases. I'm not sure what Zauker is actually useful for.
6: Halon I like. it's good to have a gas that acts like the firefighter's backpack without just space-magicing bad gases away.

marc-pelletier:
So the first three notes are actual issues and good points to make. On the lag issue, we've heard mixed things so far. Really the only way to know if the engine can handle more gases is to try. We are definitely going to have to keep an eye on that though. 

When it comes to setting up scrubbers, hopefully [this PR](https://github.com/funky-station/funky-station/pull/628) will help address that. 

As for Hyper-Noblium breaking frezon machines, that's only the case when frezon is being made at it's lowest possible efficiency thus wasting a majority of the tritium being spent. Machines like this are fine however if efficiency isn't a concern, and will have a good use in making both frezon and hypernob until we have a means of achieving temperatures below 23K. All you would have to do here is filter out the hypernob.

PR #616, opened 2025-04-07
Organ gridinv

mkanke-real:
## About the PR

Adds proper shapes and sizes to all organs and body parts as well as adding offsets to make them easier to interact with and see on the ground and in inventories.
## Why / Balance

The offsets and sizes were absolute wack before. Like a heart and a whole torso should not take up the same inventory space and the offsets were so bad the image of an item would be in a whole other gridinv space. It was pretty rough. Biggest change is you cannot put torsos in bags anymore as they have a ginormous size. There IS however a gridinv size included in case for balance reasons it needs to be lowered to Huge.
## Technical details

All YAML changes. This DOES comment out Build-a-Buddy kits for now until they can be made to work with the bigger body part sizes. Changed the ultra rare dresser spawn from human head to human heart so it fits
## Requirements

BodyBits

## Breaking changes

**Changelog**

:cl:
- tweak: All organ and body bits sizes made more sane

### DIFF
--- a/Resources/Prototypes/Body/Organs/Animal/animal.yml
+++ b/Resources/Prototypes/Body/Organs/Animal/animal.yml
@@ -68,7 +68,9 @@
     - Organ
     - Lungs
   - type: Item
-    size: Small
+    size: Normal
+    shape:
+    - 0,0,1,1
     heldPrefix: lungs
 
 - type: entity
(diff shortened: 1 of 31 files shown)


mkanke-real:
Gave it a few days for feedback on if I missed anything and gave it a second look myself. This should be ready to merge. I will start on making the build a buddy crates this week so that can be readded in a different PR

PR #647, opened 2025-04-13
Hide DoAfter for Heretic influence draining

Drywink:
## About the PR
Heretic no longer shows their DoAfter to everyone when draining influence.

## Why / Balance
It is very easy to recognize a heretic draining an influence because of the DoAfter popup that spawns. The intention is to allow heretics to be sneaky when draining influence in the open space. You can still identify a heretic draining an influence if he chooses to carry his book open to gain more points.

## Technical details
Changed the bool Hidden of the DoAfter to true.

## Requirements

**Changelog**
:cl:
- tweak: Heretic influence drain DoAfter are no longer visible to others.

### DIFF
--- a/Content.Server/_Goobstation/Heretic/EntitySystems/EldritchInfluenceSystem.cs
+++ b/Content.Server/_Goobstation/Heretic/EntitySystems/EldritchInfluenceSystem.cs
@@ -33,7 +33,10 @@ public sealed partial class EldritchInfluenceSystem : EntitySystem
         {
             MagicItemActive = ev.Handled,
         };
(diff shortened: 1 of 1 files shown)


Dot5511:
I like this change! You're supposed to be rewarded double if you use the book, making it visible you're draining influence... But normally it's visible anyways due to the bar!

Thank you for making the PR for it!

Drywink:
Just to add to your comment, even when carrying teh book, the DoAfter wil stay hidden, but if someone inspects you, they will see you carrying the book!

PR #668, opened 2025-04-19
Removes Self-Insert Plushies

DevilishMilk:
## About the PR

Removed Self-Insert plushies

## Why / Balance

They were good at the start when we were a small community to get people used to Github, but that's run its course and self-inserts are lame for a community this size now (unless everyone has one :P) but we're not that type of exclusive community

## Technical details

👎 

**Changelog**

:cl: Miiish
- remove: Killed self-insert plushies & Stowaway toys


hereelabs:
To me these things represented the people who helped build funky to where it is today, removing them feels a bit disingenuous imo

azloserbits:
Hey what if we just stop adding new plushies and keep the old ones so we honour the people who were there when the server was barely reaching 12 pop.

Ershovdi:
if a pull request is only red it probably shouldn't be merged. plushies should stay everyone loves plushies

rottenheadphones:
my plushie is just a mapped "erisia eneka" atmos plushie
You cant delete me.

DevilishMilk:
Will fix after work

PR #680, opened 2025-04-20
fixes build error in resort

DevilishMilk:
Removes PlushieSkeleton proto from the Resort map (no longer exists with #668 )

no i will not follow template


DevilishMilk:
image

PR #685, opened 2025-04-21
general BSO tweaks.

Sadie-silly:
## About the PR
Gives the Blueshield Cargo and Atmos access, gives Blueshield a CentComm radio, removes Blueshield's extra survival box.

## Why / Balance
BSO's job is to protect command and CentralCommand VIPs. Currently they can't communicate with just the other centcomm employees, and they also can't access a lot of the station where Command members could be harmed. Ex: CE gets killed in atmos, and BSO slams their fists against the glass atmos airlock, sobbing while they watch the atmosians drag the CE into the burn chamber.

## Media
{AC24ABF1-218C-4BA4-B969-1D190AE7BA84}

## Requirements

**Changelog**

:cl:
- tweak: Granted Blueshield extra access, and a centcomm overear headset.
- fix: Removed Blueshield's extra survival box.

### DIFF
--- a/Resources/Prototypes/_Goobstation/Roles/Jobs/Command/blueshield_officer.yml
+++ b/Resources/Prototypes/_Goobstation/Roles/Jobs/Command/blueshield_officer.yml
@@ -19,9 +19,18 @@
   joinNotifyCrew: true
   supervisors: job-supervisors-ntr-centcom
   canBeAntag: false
-  accessGroups:
-  - AllAccess
   access:
+  - Security
+  - Brig
+  - External
+  - Cryogenics
+  - Cargo
+  - Maintenance
+  - Engineering
+  - Atmospherics
+  - Medical
+  - Research
+  - Command
   - CentralCommand
   special:
   - !type:AddImplantSpecial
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/_Goobstation/Roles/Jobs/Command/blueshield_officer.yml
```diff
@@ -19,16 +19,9 @@
   joinNotifyCrew: true
   supervisors: job-supervisors-ntr-centcom
   canBeAntag: false
+  accessGroups:
+  - AllAccess
```
sleepyyapril, line 23: https://github.com/TheDenSS14/TheDen/blob/b0c754dc3fe6c40e3c3add1563f086f70987c58b/Resources/Prototypes/Access/misc.yml#L58
do something like this instead, they do not need access to command members' offices.

sleepyyapril (dismissed):
talked w tay, this should be changed to give them the access theyd likely need to save their vips. not chemistry, but yes to atmos. i dont think it needs anything but atmos added in all honesty.

Sadie-silly:
> talked w tay, this should be changed to give them the access theyd likely need to save their vips. not chemistry, but yes to atmos. i dont think it needs anything but atmos added in all honesty.

I think cargo access might also be a decent add, as BSO just lacks any access into cargo right now. I will make the requested changes, and also remove the extra survival box.

PR #690, opened 2025-04-22
Add support for removing whitelist via HTTP API

sleepyyapril:
yeah!


sleepyyapril:
bot is updated too but i dont think it needs to be considering i made it an optional value? shrug

PR #700, opened 2025-04-24
Cosmic Cult Revote on leader death / cryo

JoulesBerg:
## About the PR

Made it so if the cosmic cult leader fucks off they can elect a new leader and not lose the ability to 1. place the monument if it was never placed, and 2. move the monument if the leader died and its still phase 2
## Why / Balance

Right now if security finds out where the monument is, or it needs to be moved for some other reason, thats not an option.
Additionally, if they know where the monument is, and its already been moved the cult is just SOL if the leader dies, or if the ability has been used.
This way if the leader dies or cryos (or ssds I am pretty sure) a new leader can be elected, and the cult can have a chance to start anew.
## Technical details

caught the MindRemovedMessage event with an event subscription, removed the leader component from the old leader, then initiated the steward vote again. Also makes a filtered announcement to cultists that their leader has abandoned them.
After the steward vote it should set up the new leaders actions correctly based on current monument stage.
## Requirements

## Breaking changes

Right now if the leader LEAVES their body as a ghost (even if they are easily revivable) it will trigger the event and they will no longer be the leader. This could be abused, but it would be pretty obvious if it was being abused, and actionable.
This could potentially be fixed by triggering when the component is destroyed instead of when the mind component is removed, but this means you could throw the old leaders body into space and the re-vote would never trigger. 
My idea is to eventually maybe add a flavor popup to the leader who died that informs them that if they leave the body they will forsake their role.

Otherwise we'll see with play testing what might break! But I think its pretty solid.

**Changelog**

:cl: JoulesBerg
Made it so cosmic cult can elect a new leader if the old one cryos or is dead for an extended period (becomse a ghost)

### DIFF
--- a/Content.Server/_Impstation/CosmicCult/CosmicCultRuleSystem.cs
+++ b/Content.Server/_Impstation/CosmicCult/CosmicCultRuleSystem.cs
@@ -63,6 +63,7 @@ using Content.Shared._Impstation.CosmicCult;
 using Content.Server.Administration.Logs;
 using Content.Shared.Database;
 using Content.Server.CrewManifest;
+using Robust.Shared.Serialization.Manager;
 
 namespace Content.Server._Impstation.CosmicCult;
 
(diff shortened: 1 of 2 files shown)


corresp0nd:
test fail unrelated!

do you mind adding // funky begin // funky end comments to your changes? mainly cuz i dont wanna accidentally delete these edits when we port the refactor from DV (they would probably also be interested in this!!)

PR #702, opened 2025-04-24
Revs: Open Revolt!

Rainbeon:
# This PR is part of a larger Revolutionaries rework.
Any PR with this as the header is considered a part of the larger revs rework at Funky Station.

## About the PR
Open Revolt is a state where all Revolutionaries can see each other, but the Head Revolutionaries names are revealed in a global announcement by CentComm, causing them to "go loud".  This state triggers whenever a Head Rev chooses to with an action, or whenever the Emergency Shuttle fails to dock (with a ~0-20 second grace period afterward, before it occurs).

Head Revs are prevented from calling for Open Revolt within the first 10 minutes of the round, but may do so at any point afterwards.

This allows Security to progress the game state more narrowly when the revolt is triggered, allowing them to hunt down all the Head Revs specifically (instead of having to kill everyone and hope they get lucky), while also giving Revs the ability to see each other and organize together to gang up on Command without fear of being compromised by others in the late round. It also minimizes the high amount of Rev-on-Rev murder/round removal that occurs by the time fighting breaks out. 

Overall this is designed to speed up the pace of the gamemode, particularly in the late game once Revs would otherwise already be revealed, allowing both factions to focus more on actually accomplishing their goals instead of unfocused fighting without clear goals.

## Technical details
Gives every Head Rev a new action "ActionDeclareOpenRevolt" that allows them to declare Open Revolt. Open Revolt is also declared automatically when the Evac Shuttle fails to dock. Both triggers lead to the DeclareOpenRevolt() function.

Creates a new component "ShowRevolutionaryIcons" that is given to every Rev once Open Revolt is declared, and is removed by Mindshields. ShowRevolutionaryIcons is also given by HeadRevs who have ConvertGivesRevVision enabled in their component, which is also given upon declaration.

DeclareOpenRevolt() also sets the rule to it having been declared, all this manages is making sure future Head Revs that somehow appear also correctly give Rev Vision to their converts. Further, it sounds the public announcement to pending, which will play upon the next Win/Loss Condition Check (every 20 seconds, fixed). This announcement informs the general public of the situation on behalf of CentComm, and lists every HeadRev's EntityName (not mind name, to prevent temporarily SSDing to game the system) in a list to be "detained or neutralized".

## Media
https://github.com/user-attachments/assets/850a3191-6e20-4e5c-8e38-0889bd791257

https://github.com/user-attachments/assets/3a75bd32-0359-409a-8e42-d17052f38143

OpenRevolt

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added Open Revolt to Revolutionaries. Open Revolt is a state that is declared by any Head Rev, or by the Evac Shuttle failing to dock, that reveals all Revolutionaries to each other while also revealing every Head Rev's name in an announcement, allowing both Revolutionaries and Loyalist Crew to co-operate amongst themselves better in the late game.

### DIFF
--- a/Content.Server/GameTicking/Rules/Components/RevolutionaryRuleComponent.cs
+++ b/Content.Server/GameTicking/Rules/Components/RevolutionaryRuleComponent.cs
@@ -62,4 +62,10 @@ public sealed partial class RevolutionaryRuleComponent : Component
 
     [DataField, ViewVariables(VVAccess.ReadWrite)]
     public EntProtoId UplinkCurrencyId = "RevCoin";
+
+    [DataField, ViewVariables(VVAccess.ReadWrite)]
+    public bool OpenRevoltDeclared = false;
+
+    [DataField, ViewVariables(VVAccess.ReadWrite)]
+    public bool OpenRevoltAnnouncementPending = false;
 }
--- a/Content.Server/GameTicking/Rules/RevolutionaryRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/RevolutionaryRuleSystem.cs
@@ -107,7 +107,7 @@ public sealed class RevolutionaryRuleSystem : GameRuleSystem<RevolutionaryRuleCo
 
         //Add Rev Uplink
         if (!_mind.TryGetMind(traitor, out var mindId, out var mind))
-            return false;        
+            return false;
 
         var pda = _uplink.FindUplinkTarget(traitor);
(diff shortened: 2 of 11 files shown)


BombasterDS2:
I'd like to see simple dialog ui with text field that will allow head revs to make custom announcements. (psionic declaration? Headrevs just hypnotize people with flash, not using magic powers on them).

Also, don't really like idea of showing names of every head revs if someone used this. I think it's better to allow all head revs make their own "open revoult" declaration with their own custom announcement text.

Issue #681, opened 2025-04-21
General Cargo Bounty Issues


GansuLalan:
Just going to use this issue to keep track of problems I encounter with the current cargo bounties

This is what happens when people dont update tags.yml with all the tags -_-

ptrcnull:
i've also stumbled into the "Reptilian food" thing - apparently there's only a handful of allowed items: stomach, chocolate bar, nutribrick, and any kind of pizza

PR #703, opened 2025-04-24
Add GrenadeFlashBang tag to GrenadeFlashBang

ptrcnull:
otherwise doesn't get picked up by the bounty

see #681

## About the PR
Added the `GrenadeFlashBang` tag to the item of the same ID.

## Why / Balance
Because bounty items are matched by either tag or component, and the `GrenadeFlashBang` *item* did not have a corresponding type.

## Technical details
see "about the PR"

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed flashbangs not being recognized for bounties

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Throwable/grenades.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Throwable/grenades.yml
@@ -84,6 +84,9 @@
     guides:
     - Security
     - Antagonists
+  - type: Tag
+    tags:
+    - GrenadeFlashBang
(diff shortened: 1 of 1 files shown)


ptrcnull:
to be fair, i'm not sure if flash grenades shouldn't also have this tag? i haven't played enough sec to know whether there's a difference

ptrcnull:
hm, good point, i could try that

taydeo:
> It would probably be better to add the FlashBang component to the bountyCategories whitelist instead as this would help cover more cases for things that have a "flashbang" effect

Lets do this instead.

PR #708, opened 2025-04-25
Bounty prerequisites and some item clarifications 

GansuLalan:
## About the PR

Bounty items now can have prerequisite research. This should cut down on bounties being skipped because they are impossible to achieve.

There are some other minor edits like clarifications and while some bounty entries need edits @ptrcnull seems to be handling this as per #681 
## Why / Balance

currently, bounties can be generated that are impossible to complete as some lacking research is required to make the relevant item, this should be fixed now
## Technical details

when a category is selected the list of bounty items will be checked against required research if there exists any and cut down appropriately.

This can be  done via editing the researchRequired datafield on a relevant bounty item entry in the category definition, see yml for examples
## Requirements

## Breaking changes

**Changelog**

:cl:Gansu
- tweak: Certain bounty items now have prerequisite research, this will prevent them from generating until the relevant item is creatable


ReconPangolin:
I've found some new bugs that might not be fixed?

It looks like animal eyes don't count as eyes and gilded bike horns don't count as bike horns but I'll do some testing.

PR #718, opened 2025-04-28
Change Drozd and Lecter sprites to their earlier versions

The048:
## About the PR
Replaced the sprites of the Drozd and Lecter with their original sprites

## Why / Balance
The newer sprites for the Drozd and Lecter look very bad. The original sprites not only look more distinct via having more different color schemes (grey and black as opposed to both weapons having the same grey and brown scheme), but simply look more fitting. Most security items use a very clean and inorganic aesthetic, and I find it incredibly odd for the armory weapons exclusively to use an entirely different visual identity. They look more like civilian hunting weapons than space-age police firearms due to the wooden furniture, and it makes them feel disconnected from the rest of security's equipment in a way I don't think should continue.

**Changelog**
-tweak: replaced the Drozd, Lecter, and .20 rifle magazines to their original sprites
-tweak: replaced attributions for sprites in meta files for Drozd, Lecter, and .20 rifle magazines to reflect original creator.


Rainbeon:
Don't attributions need to be changed to reflect?

The048:
> Don't attributions need to be changed to reflect?

Oh. I'll fix that now.

The048:
Attributions have been changed, sorry about not getting that earlier.

PR #721, opened 2025-04-28
make plushies injectable so they leak onto you while you hold them

SpaceCat-Chan:
## About the PR
you can now inject plushies with 20u of reagents, which will slowly leak into anyone wearing or holding it (0.25u every 4 seconds, it removes 0.5u from the buffer but only applies 0.25u because i don't want this to be too useful, we have medical patches and smokes smh (this is adjustable in the component))

feel free to point out all the ways this is very stupid! :3  
i just think it's funny

## Why / Balance
this shouldn't affect balance, i just think it would be really funny to make a plushie that makes you weh or maybe a cancer plushie

## Technical details
added a LeakyObject system and component, changed all the plushies (BaseToy and any plushies which derive from it that override SolutionContainerManager) such that they have the new functionaliy.

## Media
image

## Requirements

**Changelog**

:cl:
- add: plushies are now injectable, and reagents injected will slowly and lossily pass through the skin of anyone holding the plushie.

### DIFF
--- a/Content.Server/_Funkystation/LeakyObject/LeakyObjectComponent.cs
+++ b/Content.Server/_Funkystation/LeakyObject/LeakyObjectComponent.cs
@@ -1,0 +1,19 @@
+using Content.Shared.FixedPoint;
+
+[RegisterComponent]
+public sealed partial class LeakyObjectComponent : Component
+{
+    [DataField]
+    public string SolutionName = "leaky";
+    [DataField]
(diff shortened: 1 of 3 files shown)


zoeticgrenade:
May I suggest to also make them leak some of their liquids when thrown?

SpaceCat-Chan:
with a loud thud
f7o5n9c46co81

PR #736, opened 2025-05-01
Autodoc Speed Increase

GaloGal0:
## About the PR

Increases the speed of the Autodoc.

## Why / Balance

So originally I tried to use the autodoc to replace a leg with a prosthetic - And it was painstaking. From the actual time investment in troubleshooting the operation system for the first time, to waiting for the machine to process, I was thinking "this is taking absurdly long". So I went to run some tests in a controlled environment, and the results seemed like something had to be wrong, but it was reproducible even on fresh repos.
image
These were my test results. The multitool, both at its current speed, and a slightly improved speed, were there for comparison without having to leave human error in inventory management as a factor to consider.
I don't play medical, and as such I have very limited surgery experience, even so, I was able to perform the test (removing, replacing, and closing the incision from neutral on a fresh urist) in the times provided. The same test, on the autodoc, took more than double the time. 
So, I played around with the values and believe 0.80x to be a comfortable speed - Around equal to base tools, depending on human error, but automated - which is fine, due to the higher investment required (research, construction, programming) and lack of on-the-spot versatility.

## Technical details

Tweaks the autodoc surgeryspeedmodifier from 0.50 to 0.80

## Media

https://streamable.com/jc3tag

## Requirements

- [X ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- tweak: Increased the speed of the autodoc

### DIFF
--- a/Resources/Prototypes/_Shitmed/Entities/Structures/Machines/autodoc.yml
+++ b/Resources/Prototypes/_Shitmed/Entities/Structures/Machines/autodoc.yml
@@ -79,7 +79,7 @@
     grid:
     - 0,0,3,3
   - type: SurgerySpeedModifier
-    speedModifier: 0.5 # on par with a surgeon using normal tools
+    speedModifier: 0.80 # Slightly slower than surgeon using normal tools
   - type: GuideHelp
     guides: [ Autodoc ]
 


amatwiedle:
I fully agree with this change, the speed at which the Autodoc performs surgeries currently takes forever. It's not fun to be the player undergoing a surgery with it, as you are essentially forced to do nothing while anesthetized for the long time it takes to complete a procedure.

corresp0nd (approved):

PR #753, opened 2025-05-04
Rule 11: Silicon Rules Rewrite

Rainbeon:
## About the PR
Re-writes Rule 11 to be incredibly explicit over what is and is not allowed for Silicons, as well as general definitions. They are VERY long but such is the price of Borg.

## Why / Balance
Nobody can agree on Silicon Laws, noone can even agree with what is actually written, it had to be done.

## Media
siliconRules1

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Rule 11 (Silicon Rules) has been entirely re-written.


The048:
About the definition of crew: what about syndicate IDs? they show up on the HUD but clearly indicate that the wearer isn't a member of the crew.

PR #766, opened 2025-05-07
pIED

PresentAJ:
## About the PR

I added a pipe bomb pie recipe, a pIED, that should, theoretically, add a new tool to the antag arsenal
## Why / Balance

If security doesn't power game antag players can have pipebombs on them without being discovered, who would be suspicious of a pie? It can't be cut, something I couldn't figure out how to do because it's classified as a grenade and not a food. One thing I was worried about balance was I lowered the max timer from the pipebomb's 10 seconds to 5 seconds; this was because if the target tries to eat the pIED whole, they'll have some time to throw it away. Only issues I can see this is that it can still be thrown as a normal grenade with now a 5 second timer.
## Technical details

Took a normal pie, swapped the ingredient with a pipe bomb. Also took the pipe bomb, swapped the texture with a pie. As previously mentioned, I lowered the timer from 10 to 5 seconds but kept the damage.

One thing to note is that the pie can't be sliced by any knife or sharp object, I may try that in the future but right now it's too difficult because what I originally wanted was for that to either automatically detonate it or trigger the timer but there's not established trigger code so I would have to find one out in the wild or make my own. I don't have the skills for either.

It's also not a food, it's a bomb/entity, so you can't eat it.
## Media

It's literally just plain pie. I had a video of the bomb being used but it's too big

## Requirements

## Breaking changes

**Changelog**


PresentAJ (commented):
I didn't mean to submit changes to Resources/Prototypes/Recipes/Cooking/meal_recipes.yml, if that can be redacted that'd be cool, I don't know how to remove it.

PresentAJ (commented):
Fixed the licensing and took out the construction portion because it doesn't get crafted, it's a recipe

PresentAJ (commented):
Cut the fat, condensed to fewer files, thanks mkanke

taydeo:
Is this ready?

PresentAJ:
always has been

PR #769, opened 2025-05-08
Fix issue with round ending early for zombies and ruining CC roleplay…

JoulesBerg:
… prep :(

## About the PR

Made the check for 100% conversion also only trigger if there is not an active shuttle timer.

## Why / Balance

This is a self admitted mald PR. But for a good reason. Two rounds in a row CBurn/ CC roleplay has been ruined by this logic >:( 
And I think its not intentional behavior, and I disagree with it if its intended behavior. So I am fixing it.
My reasoning, if its intended and not a bug, is that CBurn being spawned at CentComm on zombie rounds is basically standard procedure, and if there is an active shuttle timer then the zombies should be allowed to make it to CC. It will potentially make some rounds longer if there is no CBurn spawned, but I think thats a valid trade off considering that when CBurn IS spawned but this bug happens it ruins a significant time investment and is just a huge disappointment for the CBurn agents and any others involved. It also makes it not possible for someone to hide off station while waiting for evac, then try to secure evac or stealth their way onto evac. So all of their prep ALSO goes out the window. Overall I just fully 100% disagree with the current logic of "All zombies? end round immediately, no questions asked". So now it will ask the question "Well, wait a sec, is there already a round end mechanic on the way?"

## Technical details

The current logic was not checking if there was a current shuttle countdown to round end happening, so once everyone walked on the evac shuttle the round would end the next time CheckRoundEnd happened.

## Requirements

## Breaking changes

Uhh probably none from a technical stance? unless somewhere in the deep magiks (code) something else is borked.
Sometimes zombies might have to wait out a shuttle for the round to end, but to me that's a valid trade off to having legitimate role play / hard work thrown out a window.

**Changelog**

:cl: JoulesBerg
- fix: CBurn agents get to role play at centcomm now (zombie rounds should no longer end early if the shuttle has been called)

### DIFF
--- a/Content.Server/GameTicking/Rules/ZombieRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/ZombieRuleSystem.cs
@@ -140,7 +140,7 @@ public sealed class ZombieRuleSystem : GameRuleSystem<ZombieRuleComponent>
 
         // we include dead for this count because we don't want to end the round
         // when everyone gets on the shuttle.
-        if (GetInfectedFraction() >= 1) // Oops, all zombies
+        if (GetInfectedFraction() >= 1 && !_roundEnd.IsRoundEndRequested()) // Oops, all zombies
             _roundEnd.EndRound();
     }
 


JoulesBerg:
Additional comment, part of the issue could also be how the fraction is being calculated and counting zombies, Its late and I didn't look too heavily into that. Its very possible there's a better / more thorough fix. Maybe I will look into that, but I do still kind of disagree with the logic as a whole anyway so I think this is a good fix.

PR #793, opened 2025-05-16
Magistrate

taydeo:
## About the PR
Adds the Magistrate, a new legal role that is in charge of issuing warrants, executions, and going through with trials if a suspect so desires.

## Why / Balance
The Captain is too involved in decisions relating to the station when it comes to punishing criminals. The Captains primary function is to make sure all departments are running smoothly. This goal is undermined by the constant questions relating to Space Law the Captain is asked, downgrading the experience more often than not to just SecOff+. Lawyers are also too often ignored and the IAA is reserved for nagging people to do work. While a crucial part of the experience, it is not the only function they have. They are there to represent NanoTrasen as well, and by giving Lawyers, IAA, and Security Officers a new entire court to play around in, it will enhance the experience of all crew.

The Magistrate is responsible for how Space Law is enforced and has the power to change up things. For example, the Magistrate can pardon an EOC in a plea deal. Or a Magistrate can declare that a former Head Revolutionary must be contained instead of executed. Magistrate decisions are enforced by CentComm, so IAAs gotta get faxing.

The Magistrate must also preside over all trials conducted.

**Changelog**

:cl:
- add: The Magistrate has been added as a CentComm role.


ferynn:
I might suggest elaborating slightly on how magistrates are to be charged under space law, if they happen to break it. What counts as a valid charger for convicting them of space law, and the procedures etc. As they are under space law, and can be charged, but the process of actually doing that to them is a bit vague imo.

88tv:
> I'm going to be suggesting a really weird set of time requirements for Magistrate, but I think Magi should be the Hardest role to unlock, considering it is the highest RP role in the entire game.
> 
> 5 Hours NTR - This would be hit with the command/lawyer hours of IAA plus force a player to interact with the crew as the highest CC role. 5 Hours HOS - ensures you arent able to play magistrate without at least some time in the role of the person most affected by the Magi, also sec/warden time requirements (I thought they were much lower derp) 10 Hours Cap - Recycling cap's playtime requirements somewhat, but mostly for a player to have experienced multiple times what it is like to be the most important person onboard and the stress that comes with it.

I will note, it has been said in the discord prior that time requirements for Magistrate are initially going to be relatively low compared to what they're eventually planned to be. This is primarily so that we can have a lot of people playing Magistrate and trying it out to provide feedback before we lock it to what we think its requirements should objectively be.

AcornANut:
> I will note, it has been said in the discord prior that time requirements for Magistrate are initially going to be relatively low compared to what they're eventually planned to be. This is primarily so that we can have a lot of people playing Magistrate and trying it out to provide feedback before we lock it to what we think its requirements should objectively be.

I am proposing a more finalized time requirement set. but i do concur that it should not have a guy with less than like 30 hours able to play magi even with lower requirements

PR #807, opened 2025-05-21
aewofmew

misterghast:
AUUUUUUUUUUGH
idk made payments like. work. i think. so now individual accounts get like money. i think


Content.Server/_Funkystation/Payouts/PayoutSystem.cs
```diff
@@ -51,6 +51,12 @@ private void PopulateSalaryForAllJobs()
         {
             foreach (var job in salaryInfo.Roles)
             {
+                if(InitialPayoutInfo.TryGetValue(job, out var salary))
```
taydeo, line 54: There should be a HazardPay option or a Stacking option in the prototype that we can set in YAML.

taydeo (changes requested):

PR #817, opened 2025-05-24
Added Content warnings system, components, and UI functionality.

JoulesBerg:
Still need to add database functionality and either a command or a button so admins can make it so that when the user spawns the component is added to their body.

## About the PR

Added a gas for admemes.
## Why / Balance

Admemes.
## Technical details

Added the absolute bare minimum requirements for a new gas type that can be dynamically prototyped over to enable shenanigans.
## Media

image

Screenshot of me testing to make sure none of the atmos components I am aware of are broken with a prototyped version of the gas
## Requirements

## Breaking changes

It doesn't break anything EXISTING, but if, during an admeme an admin uploads a gas prototype that has a texture it can cause client crashes under certain circumstances. Just plain old errors under other circumstances. So, just keep your admin gas invisible. Not a problem. FIXING this seems like a nightmare, and maybe one day I will make the entirety of atmospherics dynamic, but today is not that day.
I tested thoroughly. 
Every component I could think of to test works fine / doesn't break in any way I could uncover.
**Changelog**

:cl: JoulesBerg
- add: Added an admin atmos shenanigans gas.


Icepicked:
Nice PR name

JoulesBerg:
> Nice PR name

I didn't even type that shit wtf :sob:

It just decided to copy it from a PR I never even pushed.

Rainbeon:
Hard-coding the exclusion makes me sad, I'd prefer if it was a field that could be re-used in like GasPrototype or something. just not hard-coding Gas == Admemium.

Content.Client/Atmos/Monitor/UI/Widgets/SensorInfo.xaml.cs
```diff
@@ -42,6 +42,9 @@ public SensorInfo(AtmosSensorData data, string address)
 
         foreach (var (gas, amount) in data.Gases)
         {
+            // don't show admemium on alarms.
+            if (gas == Gas.Admemium)
```
Rainbeon, line 46: Hard-coding the exclusion makes me sad, I'd prefer if it was a field that could be re-used in like GasPrototype or something. just not hard-coding Gas == Admemium.

Rainbeon (changes requested):

Content.Client/Atmos/Monitor/UI/Widgets/SensorInfo.xaml.cs
JoulesBerg: There's no afterserialization hooks in the gas system, so essentially no prototype values get synced to the client when you hot update a prototype (this is why you can't do sprites). If I add in a field for this then more people who use the gas for admemes will just accidentally cause client crashes and not understand why. Its even worse because in this case with the alarms and other things I hardcoded for admemium, its a 100% crash due to improper indexing and not just a possible crash like with the sprites.

JoulesBerg:
```
 FIXING this seems like a nightmare, and maybe one day I will make the entirety of atmospherics dynamic, but today is not that day.
 ```
 This is what I was talking about when I said this above.

JoulesBerg: So no. I won't do that.

JoulesBerg: To clarify about my response, I won't be doing this because GasPrototypes don't do anything. All of atmos is hardcoded anyway until a major rewrite. If that means you aren't going to merge it then thats fine. It won't get merged.

JoulesBerg:
Part of my reasoning is that for the most part there should not be ANY
hidden gases. Until theres a major rewrite of atmos theres a reasonably
hard limit to the number of gases we can have before it really does start
to become a performance issue. Having a hidden gas just makes atmospherics
tedious, someone dedicated could still filter admemium through a couple
different ways. Unless theres a reason to have it dynamic, like filtering
unlocking after a certain technology, theres no reason to code an entire
new feature on top of my minimal changes. The point of the PR was to be
minimal. Right now adding gases isn't something thats possible without a
good understanding of the atmospherics code anyway, so fleshing out the
prototype system is 100% something I am not going to invest my time in. I
understand your preference that it not be hardcoded, I even agree with it.
But its not worth the investment of anyones time.

taydeo:
I am closing this due to how this is implemented and all the unknowns about adding a new gas that is not displayed on any UI.

PR #855, opened 2025-06-02
Adds Velociraptor Cube to emagged biocube fabricators

DevilishMilk:
## About the PR

Velociraptor Cubes are now available in emagged biocube fabricators. They cost 10 sheets of plasma (5x that of abomination cubes) and 20 units of biomass.

## Why / Balance

The Dino Wars must continue.

They are still very expensive unless Botany gets a large supply of plasma, and only spawn velociraptors rather than the assorted dinocube box zookeepers and scientists can buy.

:cl: Miiish
- add: Added velociraptor cubes to emagged biocube fabricators


IronDragoon:
Very good idea imo, would love to see these used.

PR #857, opened 2025-06-03
Custom Names in Loadouts for non-Silicons

ferynn:
## About the PR
I built off the name system that got added for AI loadouts, and expanded that to work for non-silicon crew. By adding a tag to a role's entry in role_loadouts.yml, they can be added into this system. I started by just enabling it for Clowns, Mimes, and Musicians, as they are performers, but it can be done with any role if desired.

## Why / Balance
Doesn't effect balance. Just meant to let people make fun names for roles that could have a heavy usage of non-regular names. Eg. Clown.

## Technical details
There was an attempt at an override already, a section of code setting the player entity to the custom name, but by default it didn't actually work. Not with non-silicons at least. I'm pretty sure it is just being overwritten twice.

But basically I checked for the custom loadout tied name existing, and if it was non-null, it overrides the default name value in the profile. This results in modifications to the player entity name, the PDA/ID gen, and the station record entry. If there is no custom name entered on roles that have access to it enabled, it just defaults to the regular name.

There is still some bloat code existing around, but I figured I wouldn't touch what existed so far, figuring that could come later. There's no reason to have a special preferred silicon field now that it can just all be done in loadouts. But yeah.

## Media

Screenshot 2025-06-02 224056
Screenshot 2025-06-02 224119

Pictures of the custom name UI for mime and clown. I didn't make the UI element. Just repurposed the work of whoever did it for Station AI.

Glorbo

Entity name correctly on spawn.

miscCapture

Name shows as correct in other systems.

## Requirements

## Breaking changes

No changes to larger code structures. No predicted effects beyond the three files changed.

**Changelog**

:cl:
- add: You can add custom names to your loadout for Clown, Mime, and Musician.


ferynn:
oops. I forgot to uncomment the CL block. Should be visible now.

PR #870, opened 2025-06-05
Shuttleloop hotfix

Queer-Cats:
## About the PR
Just Blob blocking shuttle recall

## Why / Balance
When there are zombies and a blob on the station simultaneously, the zombie event continually tries calling the evac shuttle, while the blob continually recalls it, leading to annoying announcement spam. This situation is particularly common on Survival.

## Technical details
Just deleted the code for recalling the shuttle from blob, and removed reference to shuttle being unable to be called

## Requirements

## Breaking changes

**Changelog**
:cl:
- remove: Removed blob blocking shuttle recall


bVelliquette:
This is going to (effectively) remove the "Nuke the station" aspect of blob rounds (with or without zeds) too, since it allows crew to just evac from the blob, won't it?

PR #875, opened 2025-06-06
Syndicate Elite Hardsuit Brute Rebalance

ferynn:
## About the PR

I changed the brute damage resistances of the Syndicate Elite Hardsuit. It now has the same slash and blunt resist as the Blood-Red, and 5 percent more pierce than the blood red

## Why / Balance

Well, the Elite Hardsuit is just worse than the Blood-Red right now, unless you're planning on facing/using flamethrowers. Which kind of surprised me, as I hadn't noticed it until now. If this is intentional feel free to ignore 
this PR, but I'm assuming it is an oversight.

In game currently, the elite suit has 40 percent brute resist across the board, while the blood-red has 50 percent. In exchange the elite suit gets more fireproofing and rad resistance, but this seems like a very strange tradeoff for an "elite" suit that costs 25 more TC. I feel like most people just assume that it has more resists with reading or something, as it seems very not worth it in most cases compared to the blood red.

So, now I changed the brute resists to 50/50/55 for Blunt/Slash/Pierce respectively. About the same as the blood-red, with marginal increases in shooting resist to make it feel more elite.

## Technical details

Number tweak in .yaml prototype

## Media

Current Elite Suit:
image

Blood Red:
bloodRed

New Elite Suit:
eliteChanged

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Increased brute resist values of syndicate elite hardsuit to be better than the blood-red

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
@@ -610,9 +610,9 @@
   - type: Armor
     modifiers:
       coefficients:
-        Blunt: 0.6
-        Slash: 0.6
-        Piercing: 0.6
+        Blunt: 0.5
+        Slash: 0.5
+        Piercing: 0.45
         Heat: 0.2
         Radiation: 0.01
         Caustic: 0.5


McBosserson:
I thought its whole thing was sacrificing some kenetic protection for a ton of heat protection.

The048:
You have no IDEA how many flunkie teams all buy the elite suit and then get mulched because they don't know that the elite suit is complete trash right now. Even with this buff, it would still kinda be a bit stuck in its niche of fire and rad resistance, but I would be 100% down for at least allowing it to be on par with the bloodred.

McBosserson:
I forgot to check the slowdown. I got used it lacking any.

PR #883, opened 2025-06-08
Remove EOC guidelines

taydeo:
## About the PR
Removed EOC guidelines.

## Why / Balance
A discussion was made amongst the admin team as well as various discussions with community members that EoC definitions being a sort of document stuck between being defined IC and OOC does not serve the RP well. It forces security to think in terms of the gamemode / the antag instead of the crimes at hand. 

It did not help facilitate proper RP. It did not help guide security to do the right thing. For the most part, it guided security on sure fire ways to punish antagonists disproportionally to their crimes (i.e a syndicate thief getting perma'ed because of their uplink being open, instead of a plea deal or some other interesting RP.)

Removing this should allow crew a greater freedom to decide what happens to what antag, with round removal being reserved for the most robust of antags (35 counts of murder).

## Breaking changes

**Changelog**

:cl:
- remove: Removed Enemies of the Corporation guidelines for security.


ferynn:
I get the point with removing the main EOC part, but I think the second half of the guidebook page, which talks about types of threats that could result in lethals being authorized or even recommended is a good thing to keep, in some form or another. It's a good way for sec to understand the expected boundaries for force, and when they can escalate if they feel it's needed.

taydeo:
> I get the point with removing the main EOC part, but I think the second half of the guidebook page, which talks about types of threats that could result in lethals being authorized or even recommended is a good thing to keep, in some form or another. It's a good way for sec to understand the expected boundaries for force, and when they can escalate if they feel it's needed.

We will draft something to add to space law for next weeks update. This makes sense.

For now, this PR will be merged as is.

PR #885, opened 2025-06-08
Suitstorage Split

Queer-Cats:
## About the PR
Split suit storage into separate tank and back storage. The former holds air tanks and jetpacks, while the latter holds most other items that were previously held in suit storage, such as weapons and instruments. 

## Why / Balance
This is primarily meant for QoL and cool factor. As sec and salvage, having to choose between carrying an air tank and carrying a weapon is annoying. It's a minor buff in that it enables them (along with nukies and ERT) to effectively carry an extra item, but they could all already just carry it one hand, so it's not really a major change in my opinion. It also nerfs civilian suits somewhat, since they can no longer carry weapons on their back unless they can get their hands on an armoured vest (this does also mean that civillian hardsuits can't carry instruments, I'm working on another PR to correct that). 

## Technical details
Main change is just replacing the suitstorage flag with either tankstorage or backstorage, as appropriate. Species inventory templates were also updated to accomodate the new slot, including changes to the UI and Stripping Window. I also took the opportunity to slightly refactor the inventory templates for readability and useability. 

## Media
image
image
image
image

## Requirements

## Breaking changes
suitstorage, SUITSTORAGE, suitStorage, AllowSuitStorageComponent, and any other mentions of suitstorage besides the use for suit lockers, have been removed and should be replaced with either:
- backstorage, BACKSTORAGE, backStorage, AllowBackStorageComponent for back storage
- tankstage, TANKSTORAGE, tankStorage, AllowTankStorageComponent for tank storage

**Changelog**
:cl:
- tweak: There are now separate slots for air tanks and back stroage items, allowing combat hardsuits to carry both at the same time
- add: Gave bartender vest the backstorage slot, so bartenders can now wear their shotgun instead of just chopping it in half


ferynn:
Does this allow for two non-weapons on the back? Such as a jetpack and air tank at the same time?

Queer-Cats:
For that specific case, no, I specifically made the decision that jetpacks
go in the tank slot because they function as an air tank anyway. This can
be changed if people feel it's better that way, I just had to make the
decision which slot they go in, and it felt this was more appropriate.

Otherwise, yes, you can carry things like instruments and boomboxes in the
back storage slot, though civilian outfits mostly still don't have that
slot enabled. I'm thinking of adding another flag to separate civilian and
weapon back storage items, but I'm saving that work for another time.

McBosserson:
I think it would be cool if we could carry 2 spears, since the stored sprite is offset.

taydeo:
@vectorassembly We need a sprite for the new suit slot.

vectorassembly:
> @vectorassembly We need a sprite for the new suit slot.

Sure, but i am in uni right now if that's urgent.

taydeo:
> > @vectorassembly We need a sprite for the new suit slot.
> 
> Sure, but i am in uni right now if that's urgent.

nah just add it to your list

vectorassembly:
> @vectorassembly We need a sprite for the new suit slot.

Is that the with the letter E on it?

The048:
> i feel like on one hand this is very cool and i do like this pr but on the other hand it will kinda make double emergency oxygen tanks useless

Double emergency tanks are still 2x1 so they're great as backups or for if you're not wearing a suit/dealing with dangerous gases. And honestly as-is regular tanks are kinda useless if you do have a double emergency and just an annoyance if you don't have one.

Queer-Cats:
Yeah, emergency tanks continue to fill the same purpose they do now, as
emergency tanks. They take up less inventory space (functionally none if
you're hanging onto the emergency kit), and can be carried in a pocket,
which lets you carry a weapon without slowing you down as much as a sec
hardsuit would.

taydeo:
Was this given a new sprite?

taydeo:
real, QueerCats can you tell us which one is the new slot

Queer-Cats:
Oh, my bad. No, one with the E is the outer clothing slot, it's the one that has a belt on it. Would be called suit_storage.png in the old files, I've renamed it to back_storage.png and tank_storage.png

vectorassembly:
okay I deleted my reply because it had heck a ton of weird links.
I am going to draw a tank on the icon, that makes sense for a sprite called tank_storage.png :>

vectorassembly:
image

Queer-Cats:
Sick

taydeo:
YAML error related.

PR #908, opened 2025-06-11
Ilinge

The048:
## About the PR
Buffed Changeling armblade damage to a total of 20, buffed changeling shield health, made augmented eyesight free.

## Why / Balance
Me when I ruin an antagonist's signature weapon for no reason

In all seriousness, Changeling is supposed to be a melee-oriented antagonist, yet for some reason the armblade only does 12 damage total. This is, simply put, awful. There are makeshift weapons that can outstrip it, to the point where most changeling players don't even bother with it and instead just use said makeshift weapons. While the idea behind it right now is for it to have better limb damage in exchange for total damage, that sort of just encourages round removing victims after hollowing while also making changeling still bad in regular combat.

Buffing the Armblade's damage will encourage changeling players to actually use their signature ability, while also giving changeling more viability in combat if/when it does break out. As well, I don't think the limb damage needs reversion either, since it doesn't seem to cause trouble with combat so far as I have seen.

The changes to the shield simply make it more effective at being a buffer for a changeling's survivability. Similarly, the making of Augmented Eyesight free allows changelings to be more consistently capable at surviving encounters with security (or really just anyone with a flash).

Overall, these should make changeling more effective at surviving once found out, and more able to antagonize effectively without having to be syndie lite.

## Requirements

**Changelog**
:cl:
- tweak: Buffed changeling armblade damage to 20
- tweak: Buffed changeling shield health to 100
- tweak: Made augmented eyesight free


taydeo:
I am holding off on this PR due to the nature of changeling still.

Changeling sting should now grant Evo points instead of only absorbing. To buff lings before this change is made would be irresponsible of me at best.

This is a good change however, but it will be held until above change is made.

The048:
> I am holding off on this PR due to the nature of changeling still.
> 
> Changeling sting should now grant Evo points instead of only absorbing. To buff lings before this change is made would be irresponsible of me at best.
> 
> This is a good change however, but it will be held until above change is made.

Understandable

The048:
> I am holding off on this PR due to the nature of changeling still.
> 
> Changeling sting should now grant Evo points instead of only absorbing. To buff lings before this change is made would be irresponsible of me at best.
> 
> This is a good change however, but it will be held until above change is made.

Is this getting worked on by someone else or is this something you want me to add to this?

PR #909, opened 2025-06-12
Prevents humanoids delimbing/gibbing from low pressure

DevilishMilk:
## About the PR

Humans dead in space no longer automatically delimb after a few minutes
If they still have a lot of part damage, it is still possible for their parts to spontaneously fall off

## Why / Balance

You shouldn't gib from being in space/low pressure, this will likely reduce incidental round removal / med headache tons so people can get back in the action

## Technical details

Just multiplies part/individual limb damage taken by barotrauma by 0.1

**Changelog**

:cl: Miiish
- fix: Fixed low-pressure delimbings on corpses


Teasq:
> YAML FAIL UNRELATED!!!!!!!!!!!!!!!!!!!!!!

Nuh uh it's totally related YAML FAIL YAML FAIL

PR #934, opened 2025-06-17
Golden Toolbox Changes 

ThatOneGuy227:
## About the PR
Changed the Golden Toolbox to make it as unique as it supposidly should be for something that is one of a kind and spawns inside the stations vault. 
It now shares the same stats as a syndicate suspicious toolbox with it having a 8x4 inventory and dealing 20 blunt damage 
but unique to it is that it gets autoswing to make it just that little bit more special of a item
(witch arguably does not even mean that much given the toolbox does not have *that* fast of attack rate)

and now it changes it to be classified as command contraband, why would it be kept in the vault otherwise?

and has been added as a steal objective for thiefs 

now its truly a toolbox every tider aspires to get their hands on

## Why / Balance
I would not be supprised if a lot of people who saw the golden toolbox for the first time assumed it was something special only to be dissapointed its literally just a retexrued regular toolbox 

Alongside the vault itself not really having anything special inside of it other than the nuke, witch is not even relevant in a majority of rounds 

Wich it now being classified as command contraband it properly signals that this is a notable item, but its techinaly not specific to any member of command and its entirirely up to them how much will they care about it, some will want to be protective of it, others wont care, and some might just hand it over to a tider they deem worthy enough 

Now with the stat changes it becomes a properly special item that is worth paying attention to 

Alongside being added to the thief objective pool as it just felt a apropriate inclusion with it being a unique station item 

## Technical details
Mostly just simple .yml changes 
but also added a objectives folder and steal-target-group.flt file to Locale/en-US/Funkystation by recomendation 

## Requirements

**Changelog**

:cl: ThatOneMoon
- add: Rebalanced the golden toolbox that exists in the vault, now shares the same stats as a syndicate suspicious toolbox while also being capable of autoswing, now clasified as command contraband and has been added to the objective pool for thiefs


taydeo:
fix plox

PR #936, opened 2025-06-18
Shuttle loop fixes

Queer-Cats:
## About the PR
Reenabled Blob blocking the shuttle, modified its behaviour to work the same as Revs. Modified Zombies gamerule to just end the round if the shuttle is blocked to prevent an hour long stalemate. 

## Why / Balance
Cause the previous behaviour was fucking ass

## Technical details
Added a ShuttleDockAttemptEvent function to blob that sends the shuttle away if there's a living blob core. Created a new ShuttleDockAttemptFailedEvent event that gets called if the shuttle fails to dock. Primary use now at present is enabling Zombie gamerule to just end the game if that occurs. 

## Requirements
- [ ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).

## Breaking changes

**Changelog**
:cl:
- fix: Blobs block shuttles again! Hope you enjoyed escaping from blobs while it lasted
- tweak: Zombies rule ends the round if the shuttle is blocked for any reason (ie, Revs or Blobs) to prevent stalemates that can last for hours.


JoulesBerg:
Notably, I think this will also cause an immediate round end if the zombie rule is active and revolutionaries cause a shuttle recall.

Personally I don't think the fix is making it so if the shuttle cannot dock and zombies is active at all (even just with initial infected and no actual zombies on station yet) the round immediately ends without any additional checks.

Just my two cents.

Queer-Cats:
I mean, I broadly agree, but you're only looking at one specific batch of
edge cases in a field of edge cases. There's plenty of instances where
things fall into a stalemate and require admin intervention to end in any
reasonable time frame. And I don't see any problem with prioritising dead
players over living players? They're still players, and they're
specifically players who can't actually meaningfully interact with the game
in any way.

You're welcome to deal with each specific edge case in its own way, but I'm
prioritising fixing the actual major problem and returning it back to
intended working behaviour, and in lieu of finding, testing, and deciding
upon a solution to every single possible edge case (which i dont believe is
actually possible), I've opted to have it fail in a predictable way that
minimises the need for admin intervention.

Queer-Cats:
I specifically accounted for edge cases by making it fail to a known
condition, that's how you deal with unexpected scenarios. Your complaint
isn't that this code will lead to unexpected behaviour, it's that the
intended behaviour is not to your liking, don't confuse those. Your
suggestion is in fact what causes the shuttle loop problem because you'd
prefer the program fail to an uncertain position for the sake of gameplay.

PR #968, opened 2025-06-26
Tones Survival down a bit

DevilishMilk:
## About the PR

Set the baseline Survival event scaling to 7-20 minutes between events (for reference, basic scheduler is 7-25) rather than 4-12.

Average Chaos (which straight up divides the time between events) has been reduced to 10 from 12. This means on a typical Survival round, 2 hours in you will typically see an event being called every 0.7 to 2 minutes rather than every .3 to 1. Basically, chills it the hell out.

## Why / Balance

Survival rounds have a tendency to end the shift in about an hour and spend the rest of it floundering with zombies, blobs, and dragons. Hopefully, this should bring your average survival round closer to the hour-and-a-half to two-hour mark.

Should still have 0 problems ending it with a bang, though.

**Changelog**

:cl: Miiish
- tweak: Tweaked survival event rates down a bit.

### DIFF
--- a/Content.Server/StationEvents/RampingStationEventSchedulerSystem.cs
+++ b/Content.Server/StationEvents/RampingStationEventSchedulerSystem.cs
@@ -69,7 +69,7 @@ public sealed class RampingStationEventSchedulerSystem : GameRuleSystem<RampingS
     {
         var mod = GetChaosModifier(uid, component);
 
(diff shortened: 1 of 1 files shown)


DevilishMilk:
Mald PR btw (I liked my last Survival round but think this is necessary)

PR #973, opened 2025-06-27
Readds Self-Insert Plushies

DevilishMilk:
## About the PR

Removes Zombies from the roundstart and midround antags pool.

## Why / Balance

Zombies are very half-baked (well, just old and unmaintained really) and don't really mesh within rounds. Additionally, they derail pretty much every round they're in and are unable to be cured via, for example, a system like Virology. In lieu of an entire overhaul, this disables them from rolling. This does not remove Romerol, so there is still potential for zombie-ops and traitor gimmicks.

## Technical details

Removed Zombies from the roundstart pool as well as midround events. No more zombies that aren't player-or-admin made. This does not remove Romerol.
Adds Romerol as a 0.1% (1/1000) chance to the maints pills pool.
Removes poison damage for pending zombies, as well as any warning that you're infected. Best not get bit!

**Changelog**

:cl: Miiish
- remove: Zombies are no longer in the roundstart or midround antagonist pool.
- tweak: Romerol is now a VERY RARE maintenance pill.
- tweak: Zombie-infected humanoids no longer take poison damage.

### DIFF
--- a/Content.Shared/Zombies/PendingZombieComponent.cs
+++ b/Content.Shared/Zombies/PendingZombieComponent.cs
@@ -17,7 +17,7 @@ public sealed partial class PendingZombieComponent : Component
     {
         DamageDict = new ()
         {
-            { "Poison", 0.2 },
+            { "Poison", 0.0 }, //funky - no damage :)
         }
     };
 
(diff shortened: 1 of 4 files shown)


DevilishMilk:
Mald PR

PR #981, opened 2025-06-27
Money transfer fixes for withdrawing cash or transferring between departments.

Currot:
## About the PR

I changed how the system handled money when it deals with withdraws and transfers from department accounts.

## Why / Balance

This makes transferring money, either cash in hand or through the transfer option in the request computers work properly. This additionally stops an easy abuse where someone could cause other departments to have negative funds.

## Technical details

I've added a new function to deal with withdraws, which calls the normal function, just with the distribution removed.
I've also added a new function to deal with deposits, in the same fashion as the withdraws.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Cash withdrawals no longer steal from other departments.
- fix: Inter-department money transfers are no longer spread through all departments.


Currot:
Just went and tested it, and it seems to work fine with #972. Cash withdraws, buying and selling are working as intended.

PR #982, opened 2025-06-27
Multiplies cold temperature damage by 5

jackel234:
## About the PR
I've increased the cold damage from temperature loss across the board by about 5x

## Why / Balance
While spacing is a real threat currently, cold damage is pathetic outside of the actual slowdown it gives you, especially when compared to fire damage. This is the beginning part to a Void heretic buff I will be making a PR for after, however this PR doesn't actually affect void because Void sucks ass.

Ultimately, this PR makes the Emergency EVA less useful in a full spacing, as it protects from the pressure still, but doesn't protect from the cold of space as well as a proper hardsuit or EVA suit, as it should be. Previously, you could wear an Emergency EVA and never really notice any ill effects when in space, or a completely spaced area.

## Technical details
I edited the coldDamage value under Temperature wherever I could find it.

## Requirements

**Changelog**
:cl: amethystowo
- tweak: Cold damage is now increased by about 5x. Make sure to bundle up!

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/NPCs/animals.yml
+++ b/Resources/Prototypes/Entities/Mobs/NPCs/animals.yml
@@ -2256,7 +2256,7 @@
     specificHeat: 46
     coldDamage:
       types:
-        Cold : 0.05 #per second, scales with temperature & other constants
+        Cold : 0.25 #per second, scales with temperature & other constants
     heatDamage:
       types:
         Heat : 0.2 #per second, scales with temperature & other constants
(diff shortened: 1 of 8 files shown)


DevilishMilk:
How much does this affect full-spacing damage?

jackel234:
> How much does this affect full-spacing damage?

It makes death a bit quicker, but not tremendously faster (it's like 20 seconds from full to crit with internals on as a human)
the main thing it affects is the emergency eva, which a standard human will now get crit in 1 minute 40 seconds with it on

jackel234:
> Rip Reptilians

and felinids, which share the same stats for temp threshold and cold damage taken, oddly enough

PR #985, opened 2025-06-28
Magistrate job icon fix

vectorassembly:
image
BOO


redryryFilialBucket:
This is exactly what I thought the magistrate icon would be. A nice progression of the grey lawyer, to red iaa, and finally the blue magi.

PR #1030, opened 2025-07-02
Home-Run Bat Nerf

TheSecondLord:
## About the PR

_I promise this isn't a mald pr I am making this as someone who has USED this a few times_

- Decreased the blunt damage of the bat from 30 to 27, so it shouldn't kill unarmoured targets in 3 swings
- Massively nerfed the swing speed to 0.40 to prevent knockdown stunlocking
- Decreased the entity launching speed from 30 to 20
## Why / Balance

Full swing speed on the bat allows for effortless knockdown stunlocking, sometimes even on multiple people at once. Getting hit should only trip you for 0.1 second while you're launched (as the stuntime on hit is 0.1) but because getting back up now has a do-after, it's closer to 1.5 seconds on the floor...which was long enough to stunlock someone as long as there's a wall to hit them into, since the weapon had base swing speed and getting hit while knocked down resets the get back up timer. 

Reducing the swing speed means your target has just enough time to get up and attempt to run. You will most likely be able to line up another hit, but the delay is enough for your target to perhaps escape the corner and force you to reposition. Keep in mind this still knocks items out of your target's hands.

The small damage tweak was made because I believe the damage was intended to be 25 instead of 30, but they forgot that inheriting from the baseball bat would mean inheriting the 5 extra blunt damage while wielded. Stacked with even more blunt damage from colliding with a wall, this thing killed and gibbed incredibly quickly. The swing speed reduction obviously does a lot of the heavy lifting in reducing the time-to-kill, but if anything it might still be overtuned, so I've made this damage reduction to make sure it isn't killing unarmoured targets in 3 hits.

Speed reduction was done because I've seen this thing clip people both into and through walls, and 30 speed was incredibly fast. Hopefully this is more reasonable.

## Technical details

Yaml changes. MeleeWeapon component now has an attackRate added.

## Media

Demonstration of the bat's swing speed post-nerf. Keep in mind this poor urist obviously isn't trying to run away...

https://github.com/user-attachments/assets/0a8b337a-60af-4414-81d2-9342ee4139cc

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Reduced the home-run bat's swing speed, launch speed and slightly reduced it's damage

### DIFF
--- a/Resources/Prototypes/_DV/Entities/Objects/Weapons/Melee/home_run_bat.yml
+++ b/Resources/Prototypes/_DV/Entities/Objects/Weapons/Melee/home_run_bat.yml
@@ -10,21 +10,25 @@
   - type: MeleeWeapon
     damage:
       types:
-        Blunt: 25 # Get to WHACKIN!
-        Structural: 30
+        Blunt: 22 # Get to WHACKIN!  Funky change - 25 > 20, since this inherits from a baseball bat it gains 5 more damage while wielded,
+                  # meaning that it actually does 30 blunt. Plus collision damage, it could kill unarmoured in 3 hits. It's now 25 for real
+        Structural: 30 # Funky - it also gains 10 more structural, but that hasn't proved to be an issue yet so it's unchanged.
     soundHit:
       collection: ExplosionSmall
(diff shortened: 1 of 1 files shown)


GreyMario:
Test failure is https://github.com/funky-station/funky-station/issues/1037, don't worry

PR #1084, opened 2025-07-08
Prevent setting ID card accesses twice when spawning players in jobs

### LINKED ISSUE #1078: NTR has AA on Cluster
## Description
On the Cluster map, the Nanotrasen Representative spawns into the shift with the CentComm PDA, which has AA. Typically, they're supposed to have the Nanotrasen Representative PDA, which does _not_ have AA.

**Reproduction**
Join or start a round on Cluster with Nanotrasen Representative.

**Screenshots**
N/A

**Additional context**
This has been a problem for weeks (minimum). On one hand, it's nice to give NTR AA on a lowpop map, but it feels unintended.

GreyMario:
## About the PR
See title. When players are spawned into station jobs via roundstart or latejoin, their PDA's ID card is assigned a job, including accesses. This means we don't need to run PresetIdCard and assign the ID card a job, including accesses, because it's already done.

This is ostensibly an upstream bug and I will be filing it as such on Wizden. I won't claim to know the best way to solve this; if they have a better idea, feel free to pull that when they fix it.

_I do not know if this somehow breaks some weird edge case somewhere._ All I did was check to make sure that NTR's accesses are correct and that the game doesn't implode.

This PR fixes #1078.

## Why / Balance
This fixes an issue with the Nanotrasen Representative, who uses a CentComm PDA, of which its accesses normally don't match Nanotrasen Representative's accesses (true AA versus basic AA + command + CentComm).

Also, it's incredibly silly to set accesses _twice_ when spawning an ID card attached to a player that has a job.

## Technical details
trycomp? remcomp. it's apparently that easy.

## Requirements

**Changelog**

:cl:
- fix: Nanotrasen Representative's ID now correctly receives its intended job title and accesses instead of becoming All Access.

### DIFF
--- a/Content.Server/Station/Systems/StationSpawningSystem.cs
+++ b/Content.Server/Station/Systems/StationSpawningSystem.cs
@@ -27,6 +27,7 @@ using Robust.Shared.Prototypes;
 using Robust.Shared.Random;
 using Robust.Shared.Utility;
 using Content.Server._EinsteinEngines.Silicon.IPC; // Goobstation
+using Content.Server.Access.Components; // Funkystation
 
 namespace Content.Server.Station.Systems;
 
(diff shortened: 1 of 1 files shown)


88tv:
The issue noted here might be the cause of (and this PR may fix) https://github.com/funky-station/funky-station/issues/1078, where the NTR has AA on Cluster.

GreyMario:
This fixes #1078 and any other issues where the NTR is reported as having AA.

GreyMario:
Maybe it's better to stop PresetIdCard from executing if it finds there are already accesses on the card. Hmm.

PR #1085, opened 2025-07-08
Adds Logistics access group, used for ordering.

DevilishMilk:
## About the PR

Adds the Logistics access group, given to Cargo Techs, QM, and the HoP (for access-granting purposes.)
Only people with this access are able to approve orders.
Cherrypicked from https://github.com/impstation/imp-station-14/pull/1500, modified slightly by me because we don't have Couriers and don't really need Logistics-locked doors.
Also removes the ordering console from the Salvage shuttle, because they don't need it anymore.

## Why / Balance

Prevents everyone and their mother with Cargo access (which I believe includes the NTR, IAAs, Salvage, and other non-Cargo jobs) from ordering whatever they want with no oversight. Now you have to actually work in the department.
Salvagers? Ask your boss to order you something.

## Technical details

Adds a new access group, entirely ymlmaxxing. One line change to one c# file.

## Media

https://github.com/user-attachments/assets/62771e5c-1445-405f-a84d-86eb97ee0f04

**Changelog**

:cl: kipdotnet, Miiish
- add: Adds the Logistics access group, used to approve orders, given to the QM, Cargo Techs, and HoP.
- remove: Removes the cargo ordering computer from the Salvage shuttle.

### DIFF
--- a/Content.Shared/Access/Components/IdCardConsoleComponent.cs
+++ b/Content.Shared/Access/Components/IdCardConsoleComponent.cs
@@ -68,6 +68,7 @@ public sealed partial class IdCardConsoleComponent : Component
         "Janitor",
         "Kitchen",
         "Lawyer",
+        "Logistics", //Imp
         "Maintenance",
         "Medical",
         "Quartermaster",
--- a/Resources/Locale/en-US/_Impstation/prototypes/access/accesses.ftl
+++ b/Resources/Locale/en-US/_Impstation/prototypes/access/accesses.ftl
@@ -1,0 +1,1 @@
+id-card-access-level-logistics = Logistics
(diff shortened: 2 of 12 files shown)


DevilishMilk:
Powergaming

GreyMario:
Might want to grant Logistics to Salvage as a skeleton crew extended access, just in case?

UristMcWiki:
> Powergaming

This seems a rather moot point with the massively increased prices to guns/ammo. I do not see why it would do anything other than force salvage to yell at the QM every 3 seconds for meds and other stuff.

Teasq:
> > Powergaming
> 
> This seems a rather moot point with the massively increased prices to guns/ammo. I do not see why it would do anything other than force salvage to yell at the QM every 3 seconds for meds and other stuff.

I think the issue is salvagers buying things that inherently aren't for their department, such as insulated gloves, guns, etc.

PR #1139, opened 2025-07-14
BSO and NTR PDAs

Catazeus:
## About the PR
Adds missing PDAs + Extra

## Media
thePDAs

## Changelog
- **Added BSO PDA and ID.**
- **Added NTR PDA and ID.**
- **Added Holoflare Pen.**
 _Spawns in BSO's PDA, one shot and slow recharge._
- **Changed BSO job icon to be much more recognizable.**
- **Fixed incorrect YML.**
 _Caused Magistrate and IAA use fallback, which just so happens to be the same sprite as IAA PDA._


88tv:
this is peak solely for implementing the BSO icon that's been stuck in rsi hell for god knows how long

vectorassembly:
Hello, please, familiarise yourself with the palette, we are moving away from tg design. The shapes are definitely interesting, but the shading and colors really look like something from wizden.

PR #1206, opened 2025-07-22
Remove Elitesuit

UristMcWiki:
## About the PR

Removed the Elite suit.

## Why / Balance

Get the suit out of here

## Technical details

YAML

## Media

One comment out

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- remove: Removed the Elite Hardsuit.

PR #1212, opened 2025-07-23
Revert #875 and rename Elite Hardsuit

taydeo:
LICENSE: MIT
## About the PR
renamed da hardsuit and reverted the buff here https://github.com/funky-station/funky-station/pull/875

## Why / Balance

makes more sense for it to be called this than an "elite" hardsuit, since it has a specific niche it occupies (allows for chemists to do funny shit and alt strategies) and gets rid of the assumption that it is better for all round damage.

**Changelog**

:cl:
- tweak: The Syndicate Elite Hardsuit has been renamed to the Syndicate Aramid hardsuit.
- tweak: The Syndicate Aramid hardsuit is now weaker against bullets and melee items.
- tweak: The Blood-Red hardsuit now has greater stamina damage resistance.


Rainbeon:
Reverts #1206

PR #1243, opened 2025-07-29
add radiation production reaction for pluoxium

marc-pelletier:
## About the PR
Adds a pluoxium production reaction that uses radiation instead of tritium. Requires a minimum of 5 mols of oxygen and carbon dioxide to be present to prevent the reaction from triggering from normal breathing. The traditional method of making pluoxium with tritium still works. 

## Why / Balance
Makes it a bit less tedious to produce. 

## Technical details
Modifies the radiation system to cache rads at all coordinates that radiation sources can reach. The atmos reaction, when on a tile, then checks the cached tiles for it's coordinates, and if present gets the radiation level. This should hopefully keep checks on tiles relatively light weight. 

For pipes and canisters, a radiation receiver component is placed on them to check for rads. If the reaction isn't called for more than 5 seconds, the radiation receiver component is removed. 

## Requirements

## Breaking changes
None.

**Changelog**
:cl:
- tweak: Pluoxium can now be produced by exposing oxygen and carbon dioxide to radiation.


McBosserson:
I can't wait to see a gas chamber sticking out the side of the SM or singo.

PR #1248, opened 2025-07-29
KILL TEG KILL TEG KILL TEG

Queer-Cats:
## About the PR

Clobbered TEG effeciency.

## Why / Balance

[See here for why](https://github.com/space-wizards/space-station-14/pull/37728), however, I did not port upstream directly as it does not go nearly far enough. A mediocre TEG setup still provided enough power to support the entire station by itself. As such, I added an extra 80% power loss on the heat to energy conversion, (which, for the record, is still several times higher efficiency than irl TEGs). Testing on various roundstart TEG setups give a power output between 80 to 120 kW, which combined with solars is usually barely enough to run the station. This encourages using other engines, improving the efficiency of the TEG setups (the peak output I've managed in testing is ~200kW, I'm sure the atmos wizards can do better), or upgrading solar panels to better cope with spikes in demand without brownouts. 

## Technical details

Changed like two numbers.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: TEG has undergone a 32 fold decrease in efficiency


Queer-Cats:
@marc-pelletier Yeah, you make decent points. Problem is it's a little hard to find a sweetspot between bad TEGs still being useful and good TEGs being way too powerful. Honestly, I think the TEG is kinda just badly designed, but I don't really have any proposals for a major rework. 

I could increase the heat to power effeciency from 20% to 25 or 30%, and see how that goes for a bit. I wouldn't go much higher than 30% though, cause that's when you get TEGs producing over ~200 kW, which is the point where TEG and Solars provide all power needed forever on most stations, which then disincentivizes using the other engines.

Queer-Cats:
Honestly, perhaps Solars also deserve a nerf so we can let TEG be more powerful without the two safest (and therefore ought to be weakest) engines running most stations by themselves, but if we nerf both Solars and TEG, lowpop maps might not be able to generate enough power roundstart. 

I do think it could be interesting to just have to deal with stations that plain do not have the power to support all activities all the time, and we more or less have the tools we'd need to manage that with the new SMES/Substationg UI, but I think most people would disagree with me there.

JoulesBerg:
nerfing it to where upstream is at is probably as far as I think it should go without such a design document.

Queer-Cats:
No offence, but I fail to see how Engineering needing more to do in general
has anything to do with this. Like, yeah, engineering gameplay just being
looking after the cintainment engines is bad and boring. What's even more
boring though is engineering gameplay not even having that much to do
because Atmosians set up the TEG.

As for whether the TEG should be a main engine, I'm not a maintainer, but
the answer is pretty clear. No, it should not, at least not in its present
state. It's basically perfectly safe barring intentional sabotage (and even
then, all you can accomplish is superheating a room where the only people
who are ever in it wear fireproof clothing, andhalf the tine is open to
space anyway), it requires no maintenance, takes no resource, and isn't
that hard to set up. It's a step above Solars, but not that far ahead, and
so should not produce much more power than a solar array.

JoulesBerg:
Well, the literal maintainer for our atmos content commented already saying this level of nerf might be a bit much.
So maybe take that into consideration.

JoulesBerg:
The reason I say you should make a design document is because it enumerates those things in a way that can be reviewed and approved as part of a larger suite of changes.

On stations like Omega though, currently TEG is LITERALLY the only power source, besides some solars.
You cannot nerf it this hard without make that station almost unplayable for its purpose, which is lowpop.

There are a LOT of interconnected things to consider.

taydeo:
We will investigate a different way to address this problem.

marc-pelletier:
> No offence, but I fail to see how Engineering needing more to do in general has anything to do with this. Like, yeah, engineering gameplay just being looking after the cintainment engines is bad and boring. What's even more boring though is engineering gameplay not even having that much to do because Atmosians set up the TEG. As for whether the TEG should be a main engine, I'm not a maintainer, but the answer is pretty clear. No, it should not, at least not in its present state. It's basically perfectly safe barring intentional sabotage (and even then, all you can accomplish is superheating a room where the only people who are ever in it wear fireproof clothing, andhalf the tine is open to space anyway), it requires no maintenance, takes no resource, and isn't that hard to set up. It's a step above Solars, but not that far ahead, and so should not produce much more power than a solar array.

I do fully agree with this being an issue by the way. I think too much focus on the TEG takes away from engineers whose only real job it is to maintain power and even takes away from other things atmos could be doing. I just think this solution, while it addresses some issues, can possibly make many other issues with the TEG even worse, and will cause problems on stations where the TEG is the primary source of power.

PR #1260, opened 2025-08-01
BREATH HARPIES BREATH!!!!!!!!

BrightNibbleston:
## About the PR

I changed the air alarm to set off warning at 27.2%
## Why / Balance

I changed this so distro could be changed so harpies could breath
## Technical details

changed the "id: stationOxygen   lowerWarnAround: !type:AlarmThresholdSetting
    threshold" to 1.5

## Requirements

**Changelog**

:cl:
- tweak: Changed air alarms and such to allow harpies to breath without setting off air alarms


JoulesBerg:
Do harpies need 34% oxygen?

I thought they gasped because they were particularly sensitive to waste gases like CO2

BrightNibbleston:
> Do harpies need 34% oxygen?
> 
> I thought they gasped because they were particularly sensitive to waste gases like CO2

They dont need 34% (thats why its the upperbound, otherwise known as the danger limit)

harpies do gasp because of that but I have asked fern of functionality what amount of oxygen do harpies require to breath properly and it was 24-25%. I changed it to 27.2% to allow for some leeway as even 50% oxygen is safe within distro but logicaly doesn't make sense.

marc-pelletier:
Is this change only for the sake of Harpy's?

BrightNibbleston:
> Is this change only for the sake of Harpy's?

Not entirely but tgats the main reason. As it doesnt make sense for NT to worsen one species air when fixing it resualts in 0 problems. It also makes it so having a little bit more oxygen than needed for everyone else is not going to set off every fire lock on the station

JoulesBerg:
This is fully just a guess based on a short code dive during my lunch break.
From what I can tell, harpies breathe half as frequently, and process twice as much reagent when they do so. I only spent like 20 minutes looking at the code, so its very possible I missed something, but other than processing more gas less frequently I don't see anything that says they need MORE oxygen.
What it looks like is happening is that because they process twice as much CO2 at once as other species, they reach the thresh-hold twice as easily for the de-oxygenation to trigger, or something along those lines.

I do think that increasing oxygen in the air would theoretically counter this, since it would increase oxygenation per breath, but TBH since its a species issue I think the solution should be a species solution. Not an "every atmos has to now set 27% oxy on our server instead of the 22 everyone is used to"

I think the value of de-oxygenation that CO2 causes could either be tweaked here, "Resources/Prototypes/Reagents/gases.yml"
```
- !type:Oxygenate # carbon dioxide displaces oxygen from the bloodstream, causing asphyxiation
        conditions:
        - !type:OrganType
          type: Plant
          shouldHave: false
        - !type:OrganType #imp, thaven can breathe all gases safely
          type: Thaven
          shouldHave: false
        factor: -4
```
or the threshhold for the negative oxygenation could be tweaked (also there)

but that solution is only a softly supported guess on whats happening since I can't do a full code dive rn.

maybe give that a try and see what it does.

An alternate solution could be to give harpy lungs their own metabalizer type, which isn't a terrible idea, but also comes with drawbacks like a slight performance hit since the metabolizer types list will be longer, etc.
Not to mention its a more complicated change than tweaking that negative oxygenation function call in the yml.

BrightNibbleston:
> This is fully just a guess based on a short code dive during my lunch break. From what I can tell, harpies breathe half as frequently, and process twice as much reagent when they do so. I only spent like 20 minutes looking at the code, so its very possible I missed something, but other than processing more gas less frequently I don't see anything that says they need MORE oxygen. What it looks like is happening is that because they process twice as much CO2 at once as other species, they reach the thresh-hold twice as easily for the de-oxygenation to trigger, or something along those lines.
> 
> I do think that increasing oxygen in the air would theoretically counter this, since it would increase oxygenation per breath, but TBH since its a species issue I think the solution should be a species solution. Not an "every atmos has to now set 27% oxy on our server instead of the 22 everyone is used to"
> 
> I think the value of de-oxygenation that CO2 causes could either be tweaked here, "Resources/Prototypes/Reagents/gases.yml"
> 
> ```
> - !type:Oxygenate # carbon dioxide displaces oxygen from the bloodstream, causing asphyxiation
>         conditions:
>         - !type:OrganType
>           type: Plant
>           shouldHave: false
>         - !type:OrganType #imp, thaven can breathe all gases safely
>           type: Thaven
>           shouldHave: false
>         factor: -4
> ```
> 
> or the threshhold for the negative oxygenation could be tweaked (also there)
> 
> but that solution is only a softly supported guess on whats happening since I can't do a full code dive rn.
> 
> maybe give that a try and see what it does.
> 
> An alternate solution could be to give harpy lungs their own metabalizer type, which isn't a terrible idea, but also comes with drawbacks like a slight performance hit since the metabolizer types list will be longer, etc. Not to mention its a more complicated change than tweaking that negative oxygenation function call in the yml.


ok, so to my understanding what happens in game is harpies breath and produce enough co2 they do infact start to choke on their own exhaled co2, however. the reason why they do is because the amount of co2 takes up more of a % than the oxygen (meaning there is more co2 and less o2) which is why they start to de-oxygenate. so raising the % of o2 (and therefor increasing the mols) would allow there to be more oxygen than there is currently. which allows the harpies to breath even if the amount of co2 remains the same. (sorry if this sounds a bit repetitive or condescending, i suck at explaining stuff in detail sometimes)

an example being trit fires. 3% plasma and the rest oxygen is the norm despite 1% plasma making more trit. this is because less oxygen burns than plasma. so we end up with more oxygen than plasma which ends the burn. this is the opposite of what occurs to harpies, they "burn" using more oxygen than co2 (not literally) leaving more co2 than they can "burn" with.

I believe that changing how much air the alarms warning triggers at should be raised because even a small change can set off airlocks across the entire station and becomes a nightmare to fix and a lot of atmos mains i know of do wish the air alarms warning o2 % to be raised as can be seen within the images.

we also had a conversation about what this change may do as also seen in the pictures below

Screenshot 2025-08-05 221406
Screenshot 2025-08-05 221349

PR #1286, opened 2025-08-05
Roundstart weight tweaks

DevilishMilk:
## About the PR

Modified Secret roundstart weightings.
Completely removed Blob and Blobtraitor from the roundstart pool.
Bumped Revolutionaries from 15% to 10%
Added the remainder of these over to Traitor, changeling, and Survival.

## Why / Balance

Staff discussion from a few days ago.
For blob rounds:
Blob always feels underwhelming as an inherent roundstart gamemode due to hinging the round on one person, and works best as a midround antagonist where it serves as a diversion of time and resources to the main antagonists of the round. That's not to say it's not incredibly powerful if ignored.
Revolutionaries are my favorite gamemode, but they can be pretty exhausting and slug-festy. Lowering the chances of it rolling makes it feel more special when it does roll, IMO.

**Changelog**

:cl: Miiish
- tweak: Tweaked roundstart event weightings. TL;DR Revs down, Blobs down, Traitors, Changeling, Survival, up.

### DIFF
--- a/Resources/Prototypes/secret_weights.yml
+++ b/Resources/Prototypes/secret_weights.yml
@@ -34,14 +34,12 @@
 - type: weightedRandom # All dual antag gamemodes are goobstation
   id: Secret
   weights:
-    Traitor: 0.25 # Goobstation
-    Changeling: 0.03 # Goobstation unique antag
+    Traitor: 0.32
+    Changeling: 0.13 # Goobstation unique antag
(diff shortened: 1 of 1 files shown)


ferynn:
Is it possible to slightly increase the probability of blob rolling as a mid round, due to the removal from main round antag? I don’t know how hard a tweak that would be.

PR #1298, opened 2025-08-07
Rewrote Delta Alert

ferynn:
## About the PR

I rewrote delta alert slightly, to make it accurate as to what causes it, and pulling a lot of things from Gamma alert.  It's a mix of some of the old formatting, as well as the more fleshed out bits that are in Gamma alert.

## Why / Balance

The current form of Delta alert is actually just wrong about what triggers it and has been for a while, and that's not helpful for new people. Gamma alert is also less bad than Delta alert, but it has more 'extreme' effects on crew life and processes. Porting those bits about martial law into Delta makes sense to me.

## Technical details

Only xml changes in the the delta alert file

## Media

Screenshot 2025-08-06 181742

Screenshot 2025-08-06 181750

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Delta alert is rewritten in the guidebook. Draws from gamma alert.

### DIFF
--- a/Resources/ServerInfo/Funkystation/Guidebook/AlertLevel/DeltaAlert.xml
+++ b/Resources/ServerInfo/Funkystation/Guidebook/AlertLevel/DeltaAlert.xml
@@ -8,25 +8,31 @@ SPDX-License-Identifier: MIT
   # Station Alert - Delta
 
   <Box>
-    [italic]Called by: Malfunctioning AI Units, the supermatter delaminating, or the on-board Nuclear Fission Device being activated.[/italic]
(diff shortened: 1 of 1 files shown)


ferynn:
Yeah, that's fair. Can be shifted however you feel is best, or rewritten a different way. I was just kind of spurred into this because the current one doesn't work IC or OOC. So some kind of stopgap change felt better than keeping it as is.

ferynn:
I originally kept the style of the old delta alert, with conditionals on causes and such, but I could also write up a version with only information available to the average crew. One that's metashield appropriate. Would basically just have to scrub off a few things.

Though I guess that begs the question of would NT want average crew members to know an alert like Delta even existed?

taydeo:
A rewrite would be best then.

PR #1317, opened 2025-08-09
Port reduced motion flash effect version 3 (#37824)

hereelabs:
## About the PR
Ports https://github.com/space-wizards/space-station-14/pull/37824
"The effect is a simple black overlay with the same falloff curve as used in the shader, so it should be comparable concerning combat balance."

## Why / Balance
Flash effect can be nauseating and headache inducing

## Technical details
Original PR Edited Content.Client/Flash/FlashOverlay.cs

## Media
(Just copying the video from the original PR, but I did test and it works ingame, will add that recording if need be)

https://github.com/user-attachments/assets/9d712879-cec5-414f-a5dc-2d228c3ebeca

## Requirements

## Breaking changes

None

**Changelog**

:cl: Quantum-cross, slarticodefast
- add: The flash effect is now replaced with a reduced motion variant if the accessibility setting is enabled in the game options.


taydeo:
Add a changelog as increasing visibility in accessibility features like these is always great.

PR #1343, opened 2025-08-11
Removes MedTek cartridge from crew PDAs, boosts health analyzer usage a bit

DevilishMilk:
## About the PR

Removed the MedTek from Medical crew and the BSO.
Boosted health analyzer charge by 50%, so it should now hold about 7.5 minutes of constant use.

## Why / Balance

The MedTek seems a bit odd and out-of-place, and makes the health analyzer completely redundant. Why bother using this cool specialized tool if I have a constantly-charged better one constantly on me?

## Technical details

yml edits galore

## Breaking changes

**Changelog**

:cl: Miiish
- remove: Removed the MedTek cartridge from crew PDAs.
- add: Doctors and Paramedics now start with health analyzers.
- tweak: Boosted health analyzer charge capacity by 50%

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Items/belt.yml
+++ b/Resources/Prototypes/Catalog/Fills/Items/belt.yml
@@ -34,6 +34,7 @@
 # SPDX-FileCopyrightText: 2024 Ubaser <134914314+UbaserB@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2024 jellygato <aly.jellygato@gmail.com>
 # SPDX-FileCopyrightText: 2024 slarticodefast <161409025+slarticodefast@users.noreply.github.com>
+# SPDX-FileCopyrightText: 2025 DevilishMilk <michaellapjr@gmail.com>
 # SPDX-FileCopyrightText: 2025 IronDragoon <8961391+IronDragoon@users.noreply.github.com>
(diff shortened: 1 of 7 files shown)


PurpleTranStar:
I personally wouldve gone the opposite route, remove medical analyzers for medtek. but thats just my opinion.

LaryNevesPR:
I don't think it's a good idea...


> wont this just lead to the cmo/hop handing it out roundstart?

This is what would probably happen

PurpleTranStar:
> Cant do that cause neither of them have it anymore either
> 
> If you don't like this / would do it another way, tell me why.

Doctors are kind of worthless at the minute. with medical analyzers, anybody can heal themselves without problem. its easy to walk into medical and heal yourself without a doctor.
I feel removing the doctor's exclusive app just makes them even more worthless. Removing the medical analyzers would make it so that doctors are the only way to receive an apt analysis of the damage you've taken. The CMO will probably not hand out the medtek cartridge, because even in today's game with it, they still don't do that. I have never seen the CMO even hand out medtek to even new hires or the heads.

PurpleTranStar:
Further, further further. FURTHER comment: this could be a good segway into giving the body scanner the medical analysis instead, so you need someone else to scan your damage instead.

PR #1397, opened 2025-08-20
Disable BSO as a job

taydeo:
## About the PR
The Blueshield Officer is a strange role that has no clear place in Funky Station. This PR removes mentions of the BSO in Command SOP as well as removing the role from queue.

## Why / Balance
bso oversteps, takes away from security, makes antagonist jobs much more difficult, forces there to be a correct interpretation on how to play this role and detracts from rp a lot of the time.

**Changelog**

:cl:
- tweak: The energy revolver is now a part of the Detective load-out, and replaces the inspector.
- tweak: The energy revolver now has enough battery for 16 disabler shots, or 5 lethal shots.
- tweak: The energy revolver no longer auto-recharges itself.
- remove: Blueshield Officer is no longer a round-start role.


PWJDM:
Its.... Peak Station 14.

PR #1403, opened 2025-08-21
Glacier changes and fixes for the week of Aug 21

willowzeta:
## About the PR

The next update is gonna take a while, and the map will look substantially different the next time it is updated. Consider this the definitive pre-rework version of the map. (It won't be a planet-map version, as the codebase isn't ready for that yet.)

Glacier:
- The BSO has been reduced to atoms. A state-of-the-art tidershop has been erected where BSO's room used to be.
- The TEG is now set up to capture exhaust from the cold side by default.
- The cargo techfab has been properly mapped in.
- Removed the ore processor and recycler from salvage, since they no longer need them mapped in on station.
- The Brigmedic area now has bodybags.
- Warden now has a crew monitoring console instead of a second camera monitor. Also a banjo.
- Some pest infestation has been dealt with.
- The AI turrets should be functional now.
- The final salvager has been executed.
- hampert. He.

**Changelog**

:cl:
- tweak: Lots of fixes and changes on Glacier.


Rainbeon:
Just notifying that if I can not even begin to decipher the meaning of certain bullet points they will not be included on the dev blog

Issue #1393, opened 2025-08-19
New Job System sometimes breaks antag selection


Terkala:
Heretic selection appears to have been broken with the new job system. I'll give an example of a recent round:

Player readies up with Mime on high, CMO on low, and heretic as Yes. The game fails to put them in any role, leaves them in the lobby. It should have selected Mime/Heretic in this instance (and even if they had Heretic off, it should have spawned them as CMO). Then because it fails, it rolls randomly for other players, but since the round has already started it only managed to select command members or the station AI. Previously it would do all these checks at roundstart (and then reassign jobs based on the antag). 

Because it waits until after the round has started, it makes it so antags effectively "do not" roll first in the case where it first selects someone and has a bug.

Terkala:
It failed to assign heretic and traitor jobs to valid players. But I
haven't tested with other antags.

Terkala:
It also sometimes fails to assign latejoin thieves. Job Bartender joined late, had Thief enabled, the game failed to make them a thief. Not sure why. It did manage to eventually make a latejoin thief. Both players had multiple characters configured, and both had thief enabled, but only one of them was able to become a thief.

Terkala:
>I disagree with this, because as soon as the player session is selected to be antagonist, it will exclude roles that are not antagonist compatible.

Won't that mean that people who turn on command roles, but turn off all antags, often get left in the lobby? The previous logic would have not rolled them for the antag and thus would have left them in the queue to roll on the command jobs.

Thank you for looking into this.

Quantum-cross:
If a player only has command roles selected, then they will not even be put into the pool of people to be considered for antagonist pre-selection.

Quantum-cross:
confirmed late join antag bug, fixing

Quantum-cross:
#1421 fixes one of the two main concerns of this issue.

Quantum-cross:
I think so

Terkala:
closing

PR #1424, opened 2025-08-23
At round start -- give antagonists highest priority to get into their available roles

Quantum-cross:
## About the PR

Do a job selection pass before the normal that only considers antagonists. This has the effect of giving antagonists a higher job priority than a non-antagonist with "High"

For example, an antag with "Low priority" set to mime will have higher priority than a non-antag with "High priority" set to mime.

## Why / Balance

This will most likely fix the concern in the top post of #1393.

### Test problem example

Likely, the example described in that issue is:

Player A:
- Job Priorities: [Mime: Low]
- Character: Alice
  - Jobs: [Mime]
  - Antags: [Heretic]

Player B:
- Job Priorities: [Mime: High]
- Character: Bob
  - Jobs: [Mime]
  - Antags: [None]

### Round Start Result, Before this PR

- Player A gets selected to be a Heretic
- Job selection goes in order of priority
- Server selects Player B to be a mime, because they have it set to high priority
- Player A's job pool gets reduced to nothing
- Player A fails to spawn
- The round has no round start Heretic

Note that this can happen even if Player A has Mime set to high priority, then it would just be up to luck who gets it.

### Round Start Result, After this PR

- Player A gets selected to be a Heretic
- Job selection starts by assigning jobs to all antagonists
- Player A gets assigned Mime despite the low priority, as it's the only job they have compatible with their Heretic antagonist status
- Player B's job pool gets reduced to nothing
- Job selection continues with non-antagonists
- Player B fails to spawn
- The round starts with a round start Heretic

## Technical details

Check the comments, it's a pretty simple modification

## Requirements

## Breaking changes

**Changelog**

:cl: Quantumcross
- tweak: Improve job selection success rate with certain edge cases for antagonist players.

### DIFF
--- a/Content.Server/Station/Systems/StationJobsSystem.Roundstart.cs
+++ b/Content.Server/Station/Systems/StationJobsSystem.Roundstart.cs
@@ -12,6 +12,7 @@
 // SPDX-FileCopyrightText: 2024 Leon Friedrich <60421075+ElectroJr@users.noreply.github.com>
 // SPDX-FileCopyrightText: 2024 LordCarve <27449516+LordCarve@users.noreply.github.com>
 // SPDX-FileCopyrightText: 2025 Mish <bluscout78@yahoo.com>
+// SPDX-FileCopyrightText: 2025 Quantum-cross <7065792+Quantum-cross@users.noreply.github.com>
 // SPDX-FileCopyrightText: 2025 SlamBamActionman <83650252+SlamBamActionman@users.noreply.github.com>
 // SPDX-FileCopyrightText: 2025 taydeo <td12233a@gmail.com>
 //


Quantum-cross:
Not gonna lie I agree with you, but I was convinced that the issue of major antags not spawning round start was a bigger issue than antag players getting higher priority.

So I'm ok putting a pin in this for now. I think that this was technically an issue pre-multichar selection patch, but now that people are making more narrowly specialized characters, it might be coming out here.

Maybe I could make a patch that simply logs a high impact (I think that will put it in the text box right) admin log when an antagonist player fails to spawn. Then we can see how often this actually happens.

JoulesBerg:
it would also help narrow down whats actually happening so the fix can be more precise to address the underlying issue. I think thats a good idea.

Quantum-cross:
keep an eye on the ramifications of this 🤷

PR #1430, opened 2025-08-24
Nerfing Greencomms

PurpleTranStar:
## About the PR
This PR removes the ability to speak on the common channel through your headset. You need to use the intercom to scream for help.

## Why / Balance
The removal of greencomms changes a lot.
### Encourages in-person roleplay.
No more screaming for help at security because you broke into science and made yourself a PKA. Now, you have to actually incite a riot naturally.
### Antagonists more accessible.
Antagonists get more leeway. No longer will your sneaky stealth roleplay be shut down instantly because of a tider. You have room for error and you can more easily hide your acts. Gives more room to do stealth. Be one with the shadows.
### Reporter has a purpose.
The reporter is the main source of information. This should encourage more actual reporter roleplay and not just being a glorified tider with an announcement board. (This may require a sop rewrite..)
### We love service workers.
Service should see more usage because the bar is generally where you go to talk to people. Also a good place to go pick up a fight.
### Maints are scary.
Maints are actually scary now. Amazing, right. Be careful maints diving.
### Encourages use of alternative communications.
Faxing, holopads, and nanochat see very little useage due to greencomms just being there. Now, you have an actual reason to use them. It's your only way of communication now.

## Technical details
https://github.com/impstation/imp-station-14/commit/364aeaa798d2b060d9fbc118462462928bd71093
see here

**Changelog**
:cl:
- tweak: Common channel is now read only. You need to use an intercom to speak on it.

### DIFF
--- a/Content.Server/Radio/EntitySystems/RadioSystem.cs
+++ b/Content.Server/Radio/EntitySystems/RadioSystem.cs
@@ -20,7 +20,9 @@
 // SPDX-FileCopyrightText: 2024 Tayrtahn <tayrtahn@gmail.com>
 // SPDX-FileCopyrightText: 2024 beck-thompson <107373427+beck-thompson@users.noreply.github.com>
 // SPDX-FileCopyrightText: 2024 slarticodefast <161409025+slarticodefast@users.noreply.github.com>
+// SPDX-FileCopyrightText: 2025 Carrot <carpecarrot@gmail.com>
 // SPDX-FileCopyrightText: 2025 Currot <carpecarrot@gmail.com>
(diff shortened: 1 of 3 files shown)


McBosserson:
Over-head headsets should be whitelisted. I other words, allow command and syndicates to use their headsets.

PurpleTranStar:
> Over-head headsets should be whitelisted. I other words, allow command and syndicates to use their headsets.

With the current way this is implemented, this would be impossible.
in argument: Command has an annoucement board! Use that!

walksanatora:
Didn't goob try this and it had to be reverted cause no one liked it?

PR #1433, opened 2025-08-25
Sos app v2 wrong number

PurpleTranStar:
## About the PR
Adds an SOS app to every pda that alerts security to your presence.

## Why / Balance
With the proposal of making greencomms readonly, we need something for our lovely tiders to bug sec with. So here it is. Now you can give one standing ovation before getting round removed in maints.

Does not report your last position.

## Technical details
https://github.com/DeltaV-Station/Delta-v/pull/2812

**Changelog**
:cl:
- add: Added an SOS App that pings security.


TheSecondLord:
huh? I don't understand. 2 of the points made in the no common channel PR were "antagonists are more accessable" and "maints are scary". This directly works against them by adding something even simpler than calling for help, and literally allows you to summon sec at a press of a button. Not relaying location means nothing to anyone with the common sense to have coords turned on, especially with AI having access to both sec comms and an instant crew monitor button. I'm worried that people will start pre-opening their PDA UI before entering any possibly dangerous situations like following someone into maints, and just instantly pressing the "I'm getting attacked" button.

PR #1437, opened 2025-08-26
first aid kit slot increase, combat medkit removed from medlathe

DevilishMilk:
## About the PR

Increased first aid kit slot capacity from 4x2 to 5x2.
It still remains a 4x2 item in bags, so now you have 2 extra slots.
All first aid kits now come with a health analyzer.
Combat medkit has been removed from the medlathe

## Why / Balance

Incentivizes nicer and more sensible usage of storage containers, feels better than dumping everything in your bag being the optimal solution.
It's not TARDIS technology, it's just a visual representation of a better-organized box rather than having everything be loose.
Combat medkit has been killed from the medlathe as it's a 2x2 container with (now 5x2) units of space that cost the same amount of plastic as the base medkit.

**Changelog**

:cl: Miiish
- remove: Combat medkits are no longer available in the medlathe.
- tweak: First aid kits now have 10 slots of capacity, up from 8.
- tweak: First aid kits now all come stocked with a health analyzer.


Dragon232347:
Paramedics dispair everywhere

PR #1441, opened 2025-08-27
Pipebursting

Queer-Cats:
## About the PR

Took a crack at implementing pipebursting. In current implementation, if a pipe's pressure exceeds 9 MPa (The maximum amount that atmos devices can pump), the pipe starts leaking its contents to atmosphere; at 13.5 MPa (Higher than the operating pressures of most TEGs I surveyed while making this, both in my own tests and in real games), the pipe starts taking Structural damage, eventually destroying itself; and at 18 MPa, the pipe explodes in a similar manner to gas tanks (I genuinely could not get a pipe to hit this threshold, they tended to fracture before exploding). 

Also bundled the upstream TEG nerf, cause it's still needed. 

## Why / Balance

The TEG in its present state is simply too powerful, safe, and low maintenance, rendering all other engines obselete. My previous PRs were fairly bluntforce tools, which I do agree were kinda bad, so this is the more sensible implementation. 

From personal testing, it is trivial to operate a TEG at ~200 kW without exceeding the leak threshold - creating a TEG that does a fine job of providing supplemental power and being zero maintenance, while a TEG that was constantly leaking superheated gas into the room could sustain ~700 kW - creating a TEG that was hazardous to be around and could easily cause major problems if left unattended for too long, but is capable of powering all but the largest stations entirely by itself. 

The result is that there is a tradeoff to be had in danger vs power generation, making it a far more interesting engine to interact with, and there's inherent risk in trying to push the TEG further, but also proportional rewards in the form of greater power generation.

This also likely has some knock-on effects for other Atmos devices, especially for the HFR, but I'm not an Atmos main, so I can't really comment all that much. All that I know is that most setups I've seen probably just get somewhat less efficient and more hazardous since they generally fall in the leaky threshold, but perhaps I am mistaken in that assessment. 

## Technical details

Created a new PipeBursting Component and System. Code is largely hoisted from the GasTank explosions code, but with adjustments to make it better suited for pipes. 

## Media

image
image

## Requirements

## Breaking changes

**Changelog**

:cl: 
- add: Pipebursting!
- tweak: TEG has lower efficiency


AverageMoron:
this will render things like the HFR impossible to reasonably run and additionally will make server facets of atmos worse. I AGREE teg is overpowered but this is not a productive fix

vectorassembly:
We could introduce more pipe variants, or make them upgradable with uranium and stuff like that, to make them withstand more pressure.

AverageMoron:
> We could introduce more pipe variants, or make them upgradable with uranium and stuff like that, to make them withstand more pressure.

while that is a good idea I feel that is something that should first have a design doc made for it

vectorassembly:
> > We could introduce more pipe variants, or make them upgradable with uranium and stuff like that, to make them withstand more pressure.
> 
> 
> 
> while that is a good idea I feel that is something that should first have a design doc made for it
> 
> 

Yea, if we let pipes explode, might as well as a way to prevent that.

Dragon232347:
> > We could introduce more pipe variants, or make them upgradable with uranium and stuff like that, to make them withstand more pressure.
> 
> while that is a good idea I feel that is something that should first have a design doc made for it

Hard agree there should be a design doc for something as big as this.

taydeo:
We are considering the PR as is with the rest of the maintainers. There are some great ideas here too. I'm quite fond of the upgrade-able pipes idea.

marc-pelletier:
SS13, which by design is meant to be more tedious and harder to work with than 14, has pipe bursting that is far more forgiving than this, not cracking until the pressure within the pipe is 15,000 kpa or so higher than the pressure of the tile surrounding it, and not actually exploding until it's something like 10 times that. On top of that, most servers still have it disabled because it's horrible to work with as an atmos player.

JoulesBerg:
We had discussed pipe bursting as staff also, and determined that 50000kPA would probably be where bursting would start, and got community feedback also saying this was a good number.
Making it 9000 will make playing atmos an actual nightmare

JoulesBerg:
It would immediately make burn chambers like, not a thing. Meaning all of atmos would immediately become not a thing.

Queer-Cats:
50 MPa is a number so high the mechanic might as well not exist. Since we banned devicestacking, I've literally never seen a pipenet come close to that pressure (perhaps the HFR can achieve that, but not like anybody actually uses the thing). And for the one device that forces Atmos to interface with the rest of the Station, it does literally nothing to cap TEG power output, because at 50 fucking MPa, you're generating multiple Megawatts of power. 

To be frank, I simply think that Atmos as it currently exists is just fundamentally incompatible with station gameplay, and there does not exist a meaningful solution to fix the TEG without annoying Atmosians. The TEG has to be _substantially_ nerfed in one way or another, but there is no way to do that without making the rest of Atmos harder as well (aside from a direct power nerf, which was already rejected). 

And I just plain disagree that 9 MPa makes burn chambers tedious, it makes them not _brainless_. Instead of just setting all pumps everywhere to max, you have to actually think about injection and flow rate into and out of the chamber, and ensure your burn chambers have a way to safely and rapidly depressurise. I'm an atmos noob, and I was able to create burn chambers that could operate indefinitely and safely without oversight, I'm sure the atmos wizards can manage. 

As for the HFR, put simply, it and the TEG just can't exist in the same physical reality. I'm not even entirely sure why it was added in the first place, it doesn't really do anything interesting (if you ever see it at all).

Queer-Cats:
Okay, stepped awau from my PC to mald less, so typing from my phone now.
New pitch: keep the 9/13.5/18 threholds for regular pipes, them make tier 2
and 3 pipes from plasteel and metal hydrogen respectively that are 3 and 9
times as good (27/40.5/54 and 54/121.5/162). That way there's a significant
resource investment into getting safer pipes. So roundstart TEG will be as
I've designed, unless they want to pillage some of the plasteel to make a
more powerful design. They can alternatively use that plasteel for their
atmos setups. Then if they want to mess with the HFR, they can try risking
it with the plasteel pipes, or they can utilise the metal hydrogen pipes to
be safer (but still not entirely risk free)

Content.Server/_Funkystation/Atmos/Components/PipeBurstComponent.cs
```diff
@@ -0,0 +1,58 @@
...
+        ///     Pressure at which pipe bursts
+        /// </summary>
+        [DataField("pipeFragmentPressure"), ViewVariables(VVAccess.ReadWrite)]
+        public float PipeFragmentPressure = 4 * Atmospherics.MaxOutputPressure;
+
+        /// <summary>
+        ///     Increases explosion for each scale kPa above threshold.
+        /// </summary>
+        [DataField("pipeFragmentScale"), ViewVariables(VVAccess.ReadWrite)]
+        public float PipeFragmentScale = 2 * Atmospherics.MaxOutputPressure;
```
JoulesBerg, line 41: This is inconsistent with other gas explosion logic. The explosions will be much smaller (~90x smaller)
That's fine, but was it intentional?

Content.Server/_Funkystation/Atmos/EntitySystems/PipeBurstSystem.cs
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        public void CheckStatus(Entity<PipeBurstComponent> ent)
+        {
+            var (uid, component) = ent;
+            GasMixture? air = null;
+            PipeNode? pipe = null;
+            if (!TryComp<NodeContainerComponent>(uid, out var nodeContainer))
+                return;
+            foreach (var node in nodeContainer.Nodes.Values)
```
JoulesBerg, line 71: Using a for loop like this, you are only selecting the final node, from my testing this means only one pipe in the entire system seems to be effected by pipe bursting. and that pipe does not seem to change / update.
so if distro is full of 18kPA gas (which is not hard to do btw) only one explosion on the entire station happens, but distro stays at that pressure and doesn't explode again somewhere else.

Content.Server/_Funkystation/Atmos/EntitySystems/PipeBurstSystem.cs
```diff
@@ -0,0 +1,125 @@
...
+                air = pipeNode.Air;
+                pipe = pipeNode;
+            }
+
+            if (air == null || pipe == null)
+                return;
+
+            var pressure = air.Pressure;
+
+            if (pressure > component.PipeLeakPressure)
```
JoulesBerg, line 85: a better way to do this, since you seem to want all three of these logics to happen, would be to do a less than, and return the function early.
that way you never even bother with the following checks if the pressure is less than 9000.
```
if (pressure < component.PipeLeakPressure) return;
var environment = _atmosphereSystem.GetContainingMixture(uid, false, true);
if (environment != null)
{
    var leakedGas = air.RemoveRatio(0.25f);
    _atmosphereSystem.Merge(environment, leakedGas);
}
```
etc

Content.Server/_Funkystation/Atmos/EntitySystems/PipeBurstSystem.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+        }
+        private void UpdateMaxRange(float value)
+        {
+            _maxExplosionRange = value;
+        }
+
+        public void CheckStatus(Entity<PipeBurstComponent> ent)
+        {
+            var (uid, component) = ent;
```
JoulesBerg, line 66: I see this is also how the GasTankSystem does this, but in this case its not necessary to construct an entity before you pass it, and it just takes extra process power.
I assume the GasTankSystem does it because its handling a bunch of other shit at the same time.
You could just do `public void CheckStatus(EntityUid uid, ref PipeBurstComponent component)` or something similar and avoid a constructor call.
I am p sure you also don't need to build a separate query list up above, that's doing a second unnecessary loop over the same entities. just do
```
var query = EntityQueryEnumerator<PipeBurstComponent>();
while (query.MoveNext(out var uid, out var comp))
    {
        CheckStatus(uid, comp);
    }
```

Content.Server/_Funkystation/Atmos/EntitySystems/PipeBurstSystem.cs
JoulesBerg: to be clear, I don't understand node systems well enough to tell you what the fix for this is, but it does not work as it currently sits.

Content.Server/_Funkystation/Atmos/Components/PipeBurstComponent.cs
Queer-Cats: Yeah, I had intended to be smaller, otherwise any explosion is automatically a maxcap. Admittedly, 90 times is a lot more than I had intended when put like that, so might review that.

Queer-Cats:
Moving this to draft cause more work needs doing

JoulesBerg:
I personally think I would like to see a design doc for this (pipe bursting) and just merge in the teg efficiency nerf (one line change) as an immediate separate PR.
I agree pipe bursting should be a thing, but I think there needs to be some discussion on how it will work, how atmosians can work around it, and additional (more precise) tools for pressure control and gas movement through pipe systems that allow atmos to control pressure and prevent pipe bursting. The current way atmos works you can have a pipe jump to 60000kpa from like a burn chamber setup, and immediately go back down to 0, but it could burst in that single tick, which isn't fair to someone who by most standards is being careful but is just a victim of how atmos code works. Also, if I accidentally mix hot tritium with oxygen in a pipe it WILL burst. making atmos a lot more dangerous. A change I don't disagree with, but a discussion to be had nonetheless.

at 9000kPA as the leak threshhold and 18k as the explode threshhold with zero additional tools for atmos to stay in firm control of their pressure in the system this will immediately kill atmos on funky.

VarisciteArches:
Just to add to something if this is actually being considered,
I am really good with the HFR, I attempt to make one almost every shift I play engineering. So to actually get around the HFR pipes blowing if you intend to use it for power by connecting to the TEG, to simply just a new machine that is capable of de-pressurizing very high pressure gasses whilst keeping heat at a consistent flow.

Reason as to why not just using a gas pump is because HFR gas output loves to backup and just stop flow as gasses just pile up in that singular pipe network, and evidently leads to absolute disaster if pipebursting to be added.

Either way I personally am not in full support with pipe bursting since Atmos is perfect as it is and in my vision, doesn’t need changing.

And to add onto this, with over 600~ hours in SS13 in popular servers I have almost never seen pipebursting, Atmos was already complex enough as is. And average Atmosians already have issues making rarer gasses, the HFR just makes production of those gasses easier, so adding pipe bursting could either,

A. Make HFR completely null and useless
B. Add an upgrade to RPD that can insert certain materials in to upgrade pipes or add a way to just have RPDs upgraded with science research points without needing to consume materials to make those super reinforced pipes.
C. Just not add pipebursting at all.

Just a reminder that an average HFR could reach wayyyy beyond the limit that a reinforced pipe bursting could reach, that’s if I was reading correctly.

Terkala:
This might have issues with 9000KPA being the cap. A single volumetric pump can and will overshoot 9000KPA in practice. Even normal atmos will frequently reach 10,000-15,000KPA using a volumetric pump (simply pumping from a TEG would do it). And this would make it impossible to run most normal roundstart TEG burn chambers due to the gasses bursting the pipe if they use passive vents.

PR #1447, opened 2025-08-27
Admin Bluespace Smite

DevilishMilk:
## Media

https://github.com/user-attachments/assets/165414cb-fe8c-48fb-935d-4fa97ebb3d63
(also fixes small error in admin smite logging)

:cl: Miiish
- add: New bluespace admin smite


PurpleTranStar:
its so sad steve jobs died of ligma

PR #1448, opened 2025-08-27
Fixes recycler duplication bug

willowzeta:
## About the PR

Recyclers sometimes provide two or three times as many materials as they should yield. While I personally believed it to be intended behavior, that turned out to not be the case.

## Why / Balance

Allows for finer control of salvage's material gain. Done to help with incoming salvage changes as well.

## Technical details

Comments out a redundant Reclaim() call and removes an unnecessary math function. 

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed duplicate spawns with the recycler.

### DIFF
--- a/Content.Server/Materials/MaterialReclaimerSystem.cs
+++ b/Content.Server/Materials/MaterialReclaimerSystem.cs
@@ -18,6 +18,7 @@
 // SPDX-FileCopyrightText: 2025 Tay <td12233a@gmail.com>
 // SPDX-FileCopyrightText: 2025 pa.pecherskij <pa.pecherskij@interfax.ru>
 // SPDX-FileCopyrightText: 2025 taydeo <td12233a@gmail.com>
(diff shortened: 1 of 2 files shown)


Content.Server/Materials/MaterialReclaimerSystem.cs
```diff
@@ -185,9 +186,7 @@ public override bool TryFinishProcessItem(EntityUid uid, MaterialReclaimerCompon
         Dirty(uid, component);
 
         // scales the output if the process was interrupted.
-        var completion = 1f - Math.Clamp((float) Math.Round((active.EndTime - Timing.CurTime) / active.Duration),
-            0f,
-            1f);
+        var completion = 1f; //funkystation fix dupe bug
```
corresp0nd, line 189: does this cause any errors if the recycling process was interrupted?

willowzeta: Not to my knowledge. It's worth noting that this function doesn't even work as intended even if it wasn't unused.

PR #1485, opened 2025-09-02
speed syringe and bluespace syringe

BrightNibbleston:
## About the PR

changed the bluespace syringe to take longer to use and moved it to biochemical stasis t2 research.
changed the cryostasis syringe to speed syringe and increased its speed

## Why / Balance

I believe it would give use to these items as currently the blue space and cryostasis syringe have 0 practical use that cant be done another way that's much easier and cheaper. 

## Technical details

changed some values in specific/chemistry.yml and civilianservices.yml

## Requirements

**Changelog**

<!--
:cl:
- replaced: Cryostasis syringe with speed syringe
- tweak: bluespace syringe to be slower


corresp0nd:
speedy injects should always be Super hard to get. the current speed on the cryostasis syringe is actually a pretty good sweet spot between "being an upgrade for med" and "not insanely dangerous." also, bluespace syringes should be a t3 (even if they're generally an abandoned t3) because being able to inject that many chemicals at once is genuinely insane. 
changes like this will make em discount hypopens, possibly even better, so generally it's Not a good idea

closing cause i don't think we should merge a change like this at all, feel free to reopen if you disagree

PR #1489, opened 2025-09-04
Temporary derotation of Glacier

willowzeta:
## About the PR

Glacier is currently not being updated as I am in the middle of reworking/remaking the map. As such, I no longer consider the current version to be up to date or worthy for rotation. Glacier will be returned to rerotation (albeit under a different name a la Pioneer/Barratry) once the rework is finished and playtests show no major issues.

:cl:
- remove: Glacier derotated while I work on the revamp


willowzeta:
a cursory glance suggests the test fails are unrelated

PR #1491, opened 2025-09-04
Removed IPC torso from exosuit fabricator

willowzeta:
## About the PR
see title

## Why / Balance

I keep finding exploits with IPC torsos, and they are quite literally useless to people who aren't trying do some exploit or other.
So no harm done in removing them from the exosuit fabricator.

## Technical details

yaml

:cl:
- remove: Pulled IPC torsos from the exosuit fabricator because Shitmed continues to live up to its name


JoulesBerg:
Is it not currently possible to take a positronic from a gibbed IPC and put it into one?

PR #1494, opened 2025-09-04
Added engineering tape, an engineering holographic barrier available from the Engi-Vend 

Neoncamouflage:
## About the PR

Added engineering tape as a holoprojector that creates a holographic barrier, functionally the same as the security projector and available from Engi-Vend machines.
## Why / Balance

The Engineering SOP requires use of engineering tape despite the item not existing in the game. It also serves a useful in-game purpose of temporarily blocking off hazardous work areas and warning crew members to stay out.
## Technical details

Added engineering child entities to the holoprojector prototype
Added engineering child entities to the holosign prototype
Added retextured holoprojector sprites and a new holosign sprite
Added the new holoprojector to the Engi-Vend catalog

## Media

image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added engineering tape, a holoprojector available from the Engi-Vend. It creates a holobarrier to warn and keep crew out of dangerous worksites.


JoulesBerg:
I would just make it a passable visible sprite rather than an actual barricade. If theres physical "caution tape" around an area in a real life scenario its not even remotely impassible. It doesn't make sense for it to be a legitimate barrier like the security holobarrier, and additionally it leaves room for abuse. If I am a crew aligned engineer I don't really care about physically blocking off areas. Only about warning crew that an area could be hazardous.

mkanke-real:
I agree with the passable part but as an engi it would be annoying to cross when moving large amounts of stuff to and from a build site.

Small tweak is I would change it to engineering holotape so the naming keeps in line with the other holo projectors.

Overall though absolutely love this though

Neoncamouflage:
I agree with the feedback on making this passable, and in that case will likely change to using the holofan system for placing/removing the holosigns. Closing for now until I have time to update.

PR #1570, opened 2025-09-20
the future of gaming.

taydeo:
## About the PR
changeling is now a team antagonist

## Why / Balance
with the metashield changes changelings are shitty. this should spruce up rounds now and let changelings coorindate more effectively. since theyre an "extinct" species, it only makes sense that the last remaining members of this species would team up together.

## Technical details
new func in ChangelingSystem.Abilities.cs to add the stuff that would usually be in the shop to the changeling when they are made into one.

**Changelog**

:cl:
- add: Changelings are now a team antagonist, and are given Hivemind roundstart.


Reithauky:
So lings arent allowed to eat eachother anymore?
feels nice.

PR #1596, opened 2025-09-27
Lawyer Headset

ArionGekkota:
## Why / Balance
Lawyers start off with security comms but are given service comms in the lawdrobe. This change merges these two headsets into one, making it more convenient for lawyers to listen both to security and service comms.

## Technical details

Added a new headset with its respective sprites.
Changed the lawdrobe to stock lawyer headsets instead of service headsets
Changed the lawyer to start with a lawyer headset instead of a security headset

## Media
imagen
imagen
imagen

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Lawyers now start with a new headset that has access to the security and service channels.


taydeo:
Hi, this actually ties into a conversation that was had a lot. Do we want civilian lawyers to still have access to Security comms?

taydeo:
I agree with the channel bloat sentiment. I think removing security comms and encouraging them to use service is the right play here. It's easy to fall into the trap of over compartmentalization.

PR #1625, opened 2025-10-03
fix: Prayable Phones are now usable by atheists

### LINKED ISSUE #1566: Using red phone to call CC acts like praying at a shrine, meaning atheist characters cannot call CC
I was playing on a station alone, on Scarlet. I broke into the bridge and, on a whim, called CC using the red phone. However it only told me "You don't believe in that stuff" as if I were praying at a shrine. My character has the Atheist trait, which is responsible for this behaviour, I assume.

This is some of the relevant code:

https://github.com/funky-station/funky-station/blob/6e25cc02b3c998c5826a71d13f51894cc994390e/Content.Server/Prayer/PrayerSystem.cs#L76

The red phone call function probably uses this as well; it should be changed to not care about atheism, since CC are not gods in lore.

PileOfOxides:
## About the PR

Atheists can now use the Red Phone

## Why / Balance
Fixes #1566
Atheists probably should believe in Phones

## Technical details
Added a new Field to PrayableComponent "UsableByAtheists".
In PrayerSystem I added new logic to check for that.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Red Phone is now usable by atheists

### DIFF
--- a/Content.Server/Prayer/PrayerSystem.cs
+++ b/Content.Server/Prayer/PrayerSystem.cs
@@ -12,6 +12,7 @@
 // SPDX-FileCopyrightText: 2024 Tadeo <td12233a@gmail.com>
 // SPDX-FileCopyrightText: 2024 Tayrtahn <tayrtahn@gmail.com>
 // SPDX-FileCopyrightText: 2024 Voomra <dimon550@gmail.com>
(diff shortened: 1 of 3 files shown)


PurpleTranStar:
deny this because i think its funny

PR #1685, opened 2025-10-13
Fland mail rework

TheHolyAegis:
## About the PR

I reworked the whole mailing unit system to be more eye catching. Now more departments have a mailing unit.

## Why / Balance

It felt like the system was mostly unused, so I have moved and added more mailing units to be more visible and to cover more area.

For example, botany now has a mailing unit, so now they can send stuff directly to chemistry without needing to leave their workplace. (Chemistry could ask for medicinal plants via the holopad and botany can then send them via the mailing unit)

## Media

image
image
image

## Requirements

## Breaking changes

Zero breaking changes

## Changelog

🆑
- add: Added new mailing units to departments and some offices.
- tweak: Reworked the whole chute system using mail chutes instead of disposal chutes for easier recognition.
- tweak: Moved some of the already existing mailing units to be in a more visible location.


McBosserson:
> I think it's more ideal, the unit first was in a spot where atmos techs like never come by, where I moved it to the techs need to pass through in order to go to their workplace.

You could add an atmos mail unit, just like for chemistry and robotics. I just don't like how it's in the cramped airlock. It gets in your way in that corner, but I guess that's the point?

PR #1759, opened 2025-10-25
Make Greenshifts 50% of Rounds

ferynn:
## About the PR

I changed the weighting on Romerol from .1 to .5 

This effectively makes it go from a .1% chance to .5% chance. A five time increase, but still quite small ultimately.

I also removed .5 of weighting from the Pax/Mute/Fresium category, so that the total weight is 100 again. This makes the random chance actually what they are supposed to be, as before it was adding up to 100.1

## Why / Balance

I think the Romerol maints pill is a fun thing, and it adds a lot of interesting variation to rounds. It's meant to be rare of course, but I think making it a _little_ more common could fun. The .1 percent roll has had a very limited impact on games. It requires someone to first roll the pill, find it, and then not purify it with a chemmaster, and eat it.

It'll still be rare in this setup, but we should hopefully have a little more zombie chaos from time to time.

## Technical details

Minor changes to weighting of groups. 

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Increased Romerol strange pill chance from .1% to .5%


McBosserson:
I don't think that's how weights work.
Well, maybe in this case, since the whole weight pool adds up yo 100 before romerol.

PR #1797, opened 2025-10-29
Criminal records code rework, new criminal statuses

ilyamikcoder:
LICENSE: MIT

## About the PR
Allow for creation of new criminal statuses via prototypes.
Adds Search and Incapacitated criminal statuses.

Initially this was supposed to be a port of https://github.com/space-wizards/space-station-14/pull/36244, but because I redid basically the entire PR, and then was also asked to do a bit of a refactor, there is kind of a Ship of Theseus situation going on, so I am not sure if this counts as a port any longer. I still kept the credit intact just in case.

## Why / Balance
With how chaotic an officer's life can be, more organization cannot be harmful.

Separate status for Search is beneficial, as it is more noticeable than a Suspected status, which is usually used to denote people in need of a search.

Incapacitated status is intended to be used in situations where a criminal (or some other crew member) was removed from station's regular operations in a non-incarceration way (executed, marooned, exfiltrated, kidnapped etc.)

## Technical details

This was a port one day, but now it kind of isn't (see above).

The system works on prototypes now. I am not quite sure how to comment on this further, my brain is kind of melting at this point.

## Media
crimrecords_new_demo1
<br>
crimrecords_new_demo2
<br>
crimrecords_new_demo3

## Requirements

## Breaking changes
Everything security status-related now takes SecurityStatusPrototype instead of legacy SecurityStatus (which was just an enum).

**Changelog**

:cl: ilyamikcoder, B-Kirill
- add: Search and Incapacitated criminal statuses


ilyamikcoder:
I feel like marking someone as Eliminated when they are not, in fact, dead, to make them "kill on sight" would be quite an LRP action in itself, as that is clearly not what the status is intended to represent. The status isn't really named "Eliminate" after all...

That being said, naming it something more soft like "Deceased" could certainly be an option. Do tell what you think.

(As for "Hostile", the [wizden PR](https://github.com/space-wizards/space-station-14/pull/36244) probably explains it better - it's a separate mark for wanted people who are _very dangerous_ and should be apprehended with preparation for self-defense. It's not intended to mark people who are just acting hostile but not guilty for anything. Admittedly, if the concept isn't clear enough with a name like that, I am not entirely sure how I would name it otherwise in a way that wouldn't be too lengthy. Should such a distinction not be desired, I could certainly cut out - it might as well be not necessary as it was pulled from an LRP codebase.)

vectorassembly:
I like the search one.

Content.Server/CriminalRecords/Systems/CriminalRecordsConsoleSystem.cs
```diff
@@ -158,18 +162,30 @@ private void OnChangeStatus(Entity<CriminalRecordsConsoleComponent> ent, ref Cri
...
             // released on parole
             (_, SecurityStatus.Paroled) => "paroled",
             // prisoner did their time
             (_, SecurityStatus.Discharged) => "released",
             // going from any other state to wanted, AOS or prisonbreak / lazy secoff never set them to released and they reoffended
             (_, SecurityStatus.Wanted) => "wanted",
+            // person no longer deemed hostile
+            (SecurityStatus.Hostile, SecurityStatus.None) => "not-hostile",
+            // person returned from dead... somehow
+            (SecurityStatus.Eliminated, SecurityStatus.None) => "not-eliminated",
```
sleepyyapril: oh my gods

sleepyyapril:
im blaming wizden

PR #1800, opened 2025-10-30
Enabling powered prying for the fireaxe

pirakaplant:
LICENSE: AGPL AND MIT
## About the PR

The fireaxe (as well as the fire axe* and the metal hydrogen axe) can now pry open powered doors.

\* "Fireaxe" refers to the Nanotrasen axe found in atmospherics and the bridge. "Fire axe" refers to the Syndicate axe that can be purchased through the uplink.

## Why / Balance

The reason Atmospherics has a fireaxe in the first place is part of their job identity. They're firefighters in addition to managing distro and the TEG, so they should be well-equipped to rescue people from fires.

The problem with the current fireaxe is, it's not very good at that. If someone is trapped in an department with a fire or other atmospheric hazard, the existing fireaxe is not going to help with that unless atmos goes and manually depowers all the doors just to get someone out. And at that point, why not just use a crowbar?

As for balance, this shouldn't be too much of a problem, especially when you compare it to the jaws of life. It's not as good at prying non-powered objects as the jaws of life, and lacks the jaws' wirecutter function. It can pry hull plating, but overall it's still a downgrade. Therefore, it feels more than fair for atmos to share this when the CE gets a strictly better version for themselves in their locker.

## Technical details

The fireaxe now has `pryPowered` set to `true`. Since the fire axe and the hydrogen metal axe inherit from the fireaxe, this change also affects them (which is intended).

## Media

N/A.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The fireaxe, fire axe, and metal hydrogen axe can now all pry open powered doors.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Melee/fireaxe.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Melee/fireaxe.yml
@@ -30,6 +30,9 @@
 # SPDX-FileCopyrightText: 2024 Winkarst <74284083+Winkarst-cpu@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2024 metalgearsloth <31366439+metalgearsloth@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2024 nikthechampiongr <32041239+nikthechampiongr@users.noreply.github.com>
+# SPDX-FileCopyrightText: 2025 QueerCats <jansencheng3@gmail.com>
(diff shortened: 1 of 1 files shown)


ferynn:
The point of the fire axe isn’t to pry doors open in a crisis, it’s to smash them open. At least in my view.

YaraaraY:
I like the change. A good alternative, though, if we don't want to give Atmos pseudo-jaws, is to up the structural damage of the fireaxe to make ripping down airlocks faster. Right now you're usually better off just RCDing the doors.

PR #1817, opened 2025-11-01
[TG Sprite Port] TG Sprite Funkification, part one!

YaraaraY:
## About the PR
Ported and edited sprites from TG to replace some of our uglier upstream sprites, this time focusing on outer suits that are commonly worn and seen. And biosuits. And the paramedic jumpsuit, to match the voidsuit I made.
- Atmos Firesuit
- Firesuit
- Emergency EVA
- Spacesuit
- Janitor, virology, generic, security, science Biosuits
- Radsuit
- Paramedic Jumpsuit

I will port more sprites later! But this is the first batch.

## Why / Balance
They look much nicer than the old yucky sprites from upstream.

## Technical details
Most of the editing was recolouring the suits to our Splendor Station palette. There is still some work to be done, mainly on the Vox variations of these suits. I replaced the .PNGs directly in the .rsi folders.

## Media
emergency suits
eva suits
atmos suits
firesuits
radsuits
biosuits
paramed

## Requirements

## Breaking changes

**Changelog**
:cl: AraiMaia
- add: New sprites for: Firesuits, Atmos Firesuits, Radsuits, Biosuits, Emergency EVA, EVA suits


Dragon232347:
Holy peak

Teasq:
Hiya! I've got a few thoughts:

Firstly, I think the biosuits are really cute. I love the gas mask and for once department colours on an outfit actually makes sense. Both of the atmos firesuits are cute too, but I know vector has a really poor opinion of transparent sprites in general so that might be seen as a problem. It doesn't use the palette though.

I really like the inflatable look of the HEV suit, but there's some inconsistencies between the back of the legs not lining up with the side view. It also doesn't use the palette.

The emergency EVA I think is fine. It's just a better EVA suit, but it needs to use the palette. As for the normal EVA, I think it could be made a little darker overall as rn it feels really overexposed. It's also slightly off of the palette.

As for the paramedic outfit, I think it looks great but it also doesn't use the palette at all. I like the little hi-vis strips and the cargo pant pockets though.

The Vox also seems to have a super rough-looking biosuit which I'm certain wasn't intended and probably missed:
image
I colour picked nearly all of the sprites too and they all seem very slightly off-palette. I'm unsure if that was intended.

Overall, I think these are fine as temporary replacements for the time being but I don't really like how they're ports of another server. I think Funky resprites in general have been trying to create their own look separate from other servers and I don't think ports from other servers as permanent additions are constructive towards that.

I'd appreciate if the title maybe had something along the lines of [TG Sprite Port] or something similar to make it easier to find down the line for Funkification.

PR #1852, opened 2025-11-04
IAA administrative glasses instead of security glasses

TheHolyAegis:
Removed the security glasses loadout group, and added the administrative glasses permanently to the eye slot

## About the PR

I removed the security glasses loadout group from the IAA, to instead add the administrative glasses to their loadout.

## Why / Balance

IAA SoP mentions that the IAA is allowed to carry their flash and administrative glasses, not security glasses. The IAA was technically carrying the security glasses illegally.

## Technical details

No code changes

## Media

image

No red glint, since it's now administrative glasses

## Requirements

## Breaking changes

No breaking changes

## Changelog

🆑
- remove: Removed the security glasses loadout group from the Internal Affairs Agent.
- add: Instead, added administrative glasses to their default loadout


Idem778:
So I’ll be honest, sec glasses are really useful for IAAs because they have to go into the sec and brig (which sometimes have a flasher) quite a lot, now the actual reason is of course the sec hud. IAAs really benefit from the sec hud because it can see mindshields and wanted levels. And the why is just kinda bad… why not just fix the SoP?

PR #1879, opened 2025-11-07
remove rattler from uplink and buff wyvern

Mori428:
## About the PR

removed rattler from uplink and increased wyvern fire rate to 2.4
## Why / Balance

rattler is far too powerful and effectively invalidates the python by being a better revolver and I might rework it when I feel like it, and as for the wyvern I feel its current state is far too weak and buffing it would put it in a good spot balance wise I believe
## Technical details

changed a few values and commented out the rattler so it does show up in the uplink
## Media

im pretty sure media isnt needed
## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: removed rattler from uplink
- tweak: buffed wyvern to 2.4 fire rate


Mori428:
> "wyvern being too weak" as 50 ae does 38 damage

it has low fire rate and bad recoil the bullets do good damage but if you can only reliably hit shots in close range while having a lower dps than most guns then it really isnt that effective especially considering that its meant to be a gun only found in the vault and 40 tc in the uplink

JohnnyDangerSpace:
john skill issue pr

PR #1923, opened 2025-11-10
Port Entropic Colossus from DeltaV

NoElkaTheGod:
## About the PR
Ported the Entropic Colossus antagonist from DeltaV.
Colossus can either spawn as a midround antag (about as rare as a dragon, and pretty similar to one in behaviour), or be created by a cosmic cult when using the "Null Fragmentation" ability on a borg.
Colossus is very tanky (400 HP and good resistances) and hits like a truck (60 damage per swing but 0.4 attack speed), but is very slow (base running speed of 3 compared to 4.5 of an average humanoid) and clumsy (cannot interact with things aside from dragging).

It's unique abilities include:
- Colossal ingress. Allows the colossus to pry any airlock, even a bolted one. Yeah, that's it.
- Slumber shell. Stuns the colossus for 30 seconds, after which heals it for half of it's current damage. Needless to say, if one uses this mid-combat, it's very much dead. However, it can be fairly strong if used carefuly.
- Entropic Sunder. After using it on an empty tile, the colossus vanishes, and the selected spot is indicated. After 1.5 seconds, the colossus appears again at the selected spot, slamming it's fist into the ground, which sends a damaging shockwave around it. Requires direct line of sight (doesn't work through windows)
- Beckon an Effigy. This one is only available to the midround spawn. It must be used near a specific location, and requires a 3x3 tiles of empty space. When used, it spawns an Effigy of Entropy (which currently is just a meat anomaly with extra steps, but I will be reworking it in the future). Doing so is the primary objective of the midround colossus.

When a cosmic cultist uses Null Fragmentation on an unsuspecting cyborg, the target is trapped in a Vacuous Chantry, immediately alerting the entire station of it's location. If the chantry survives for 150 seconds, the cyborg is transformed into a colossus. However, if the chantry is destroyed before that, the cyborg is freed safely.

## Why / Balance
Brings more variety to cosmic cult specificaly and antags as a whole. Allows coscult to actually interact with borgs. Brings Null Fragmentation to completion. And most importantly, looks sick as hell.

Balance-wise it's pretty much as it was over on DeltaV, slightly tuned down in some places and polished in others. Notably, the sunder+punch one-shot combo is still technically possible, but should be easy enough to avoid to count as a skill issue.

## Technical details
A whole lot of cherry-picks with minor tweaks. Did my best to remove any features not related to the colossus, as well as to comment on all the Funky-specific changes.

## Media

The Colossus

изображение

Eeping

изображение

Entropic sunder in action (my PC is ass and so is the recording).

https://github.com/user-attachments/assets/9bcc1324-47f9-43f6-997b-411cea972309

The lawset (known as cultimov), and available radio channels (actually can't talk on common).

изображение

A guidebook entry is also included!

изображение

## Requirements

**Changelog** 
🆑 NoElkaTheGod, AftrLite
- add: New antagonist, Entropic Colossus, has been added. It can appear mid-round, or be created by a cosmic cult. Make sure to read the guidebook!

### DIFF
--- a/Resources/Prototypes/_DV/CosmicCult/Mobs/colossus.yml
+++ b/Resources/Prototypes/_DV/CosmicCult/Mobs/colossus.yml
@@ -286,7 +286,7 @@
     nameSegments:
     - ColossusNames
     - ColossusTitlesBase
-    nameFormat: name-format-colossus
+    #nameFormat: name-format-colossus # Funky, tis doesn't work
   - type: MobThresholds
     thresholds:
       0: Alive


NoElkaTheGod:
From my testing, I conclude that this is in a "good enough" state. Ready for review.

PR #1937, opened 2025-11-11
Shotgun Breaching Damage + Energy Revolver Charge Buff

Xcybitt:
## About the PR
Made buckshot have the same structural damage as wizden, and set slugs so that it would break an airlock in a full bulldog mag. Also halfed the battery and usage of the energy revolver, or in other words just made it charge faster.

## Why / Balance
I feel that with the bulldog having more price vs the C-20, this would give it a reason to be used more. Also breaching with shotguns is just badass. Slugs are that since I feel having an option to long range breach is interesting. For the revolver, I feel needing two minutes with a normal recharger to 100 from 0 just be a tad bit much.

## Technical details
YAML

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Added structural damage to shotgun slugs and pellets.
- tweak: Doubled the charge rate of the energy revolver.

### DIFF
--- a/Resources/Prototypes/_Goobstation/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
+++ b/Resources/Prototypes/_Goobstation/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
@@ -116,6 +116,6 @@
     damage:
       types:
         Blunt: 10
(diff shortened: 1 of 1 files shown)


Xcybitt:
Alr, I'll go unnerf it when I got the time

PR #1964, opened 2025-11-13
Role fixes

TheHolyAegis:
## About the PR

Added a translation to the time tracker of the CCI, the absence of it prevented it from displaying in the stats list.

I also removed both the CCI and the Space Cowboy from the lobby screen (their icons).

## Why / Balance

It was weird that the CC intern wasn't displayed in the stats, since ERT ghostroles are. So now it actually displays it.

Also removed the CCI and Space Cowboy icons from the lobby screen, since they are ghostroles, and shouldn't be displayed in the lobby.

## Technical details

The Central Command Intern stat wasn't displaying because it's translation was missing, which is now fixed (it should display the time played before this update).

I changed the setPreference to false for the CCI and Space Cowboy, their icons now won't show up anymore (like any other ghostrole, like any ERT ghostrole)

## Media

image

No more CCI and Space Cowboy icons

image

(I did in fact play this much CCI.... just kidding)

## Requirements

## Breaking changes

It shouldn't break anything, it's just "yet another markup language" (YAML/.yml) changes

**Changelog**

🆑
- add: Added the Central Command Intern playtime to the stats, which wasn't displayed before due to a missing translation
- remove: Removed both the Central Command Intern and the Space Cowboy icons from the lobby screen, which are ghostroles and shouldn't be displayed in the lobby.

### DIFF
--- a/Resources/Prototypes/_Funkystation/Roles/play_time_trackers.yml
+++ b/Resources/Prototypes/_Funkystation/Roles/play_time_trackers.yml
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: 2025 Josh Hilsberg <thejoulesberg@gmail.com>
 # SPDX-FileCopyrightText: 2025 JoulesBerg <104539820+JoulesBerg@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2025 Tay <td12233a@gmail.com>
+# SPDX-FileCopyrightText: 2025 TheHolyAegis <sanderkamphuis719@gmail.com>
 # SPDX-FileCopyrightText: 2025 kbarkevich <24629810+kbarkevich@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2025 mkanke-real <mikekanke@gmail.com>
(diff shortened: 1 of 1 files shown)


mkanke-real:
Thank you for the space cowboy fix! I had been meaning to do it but havent had time.

Resources/Locale/en-US/_Funkystation/job/job-names.ftl
```diff
@@ -8,3 +8,7 @@ job-name-internal-affairs-agent = Internal Affairs Agent
 job-name-magistrate = Magistrate
 job-name-central-command-intern = Central Command Intern
 job-name-space-cowboy = Space Cowboy
+
+# Role timers - Make these alphabetical or I eat you
+
+CentralCommandIntern = Central Command Intern
```
corresp0nd: make this follow the same naming rules as the rest of the ftl strings

corresp0nd (changes requested):

TheHolyAegis:
Should I push the changes or keep it with the wrong convention?

PR #1978, opened 2025-11-14
New decals for mappers

willowzeta:
## About the PR

Added two variants of a F.N.K.Y. Sector sign to replace the generic Space Station 14 signs, and a set of singleton floor tiles for each job to build navigation guides.

## Media

**Changelog**

:cl:
- add: Added some new decals for mappers to play with.


JohnnyDangerSpace:
The like maybe 2? mappers will heavily enjoy this

PR #2003, opened 2025-11-16
Players can take actions while prone again. Percent chance to hit prone targets while standing.

ferynn:
## About the PR

Reverted PR 993, which removed the ability to take actions while prone. This was in response to views about the combat meta being too oriented towards being prone.

I do think it has negative impacts on the game though, as removing all prone actions hit a _lot_ more than just guns.

With that in mind, I readded the ability to take actions while prone. I added a percent chance for standing targets to hit prone targets with projectiles.

## Why / Balance

Let's people do things while prone again, and remove all the many annoyances that were caused by that.

Currently there is a 20 percent chance for projectiles fired by standing targets to hit prone targets in their line of fire. This means being prone isn't total immunity, especially against automatic weapons. Stealthed targets will eventually be revealed by their bleeding/damage reactions from sprayed bullets.

Also, as a reminder. If you are prone, you hit other prone people 100 percent of the time. This is not a new addition, but an existing mechanic. So if an enemy is prone, or perhaps prone and invisible, you can drop yourself and hit them.

## Technical details

Added a initialization of the Random class in the class that determines projectile non-collision. When a projectile would potentially hit a prone target, it generates a new random number and compares it to a given threshold. Currently it is .2, and the rand value is from 0-1. 

## Media

N/A

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Players can once again take actions while prone
- add: Shooting projectiles over a prone target has a 20 percent chance of hitting them. (Firing while prone always hits prone targets)

### DIFF
--- a/Content.Shared/Cuffs/SharedCuffableSystem.cs
+++ b/Content.Shared/Cuffs/SharedCuffableSystem.cs
@@ -33,6 +33,7 @@
 // SPDX-FileCopyrightText: 2024 themias <89101928+themias@users.noreply.github.com>
 // SPDX-FileCopyrightText: 2025 JoulesBerg <104539820+JoulesBerg@users.noreply.github.com>
 // SPDX-FileCopyrightText: 2025 Tay <td12233a@gmail.com>
+// SPDX-FileCopyrightText: 2025 ferynn <witchy.girl.me@gmail.com>
 // SPDX-FileCopyrightText: 2025 pa.pecherskij <pa.pecherskij@interfax.ru>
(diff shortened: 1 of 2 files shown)


mkanke-real:
Unless this addresses people being able to to go prone, start and keep moving while shooting then we are going to fall right back into the crawl combat meta. Unfortunately 20% is still a huge boost to whoever goes prones combat ability if they can still move while shooting.

taydeo:
I think with the prediction PR shooting should feel a lot better, especially with trying to hit people who are prone.

Dragon232347:
FUCKING FINALLY

PR #2018, opened 2025-11-18
Revert "Bodycams are fabbable, camera monitor has a speaker, det gets listening radios"

taydeo:
Reverts funky-station/funky-station#2012

crashes server


JohnnyDangerSpace:
local chud host fails to playtest pr, embarrassingly has to revert change.

PR #2025, opened 2025-11-18
Changeling Sting is Live

Xcybitt:
## About the PR
Uncommented out the changeling infect sting, hopefully with the new command metashield you CAN get cured if you show up with symptoms, but if not, you can get deconverted. With this, the hive can grow.

## Why / Balance
Genuinely, other that THE incident, why was this an issue other than making changelings a conversion antag? We already made em become a team antag, so leaving an option for becoming conversion antags sounds interesting. At least with this, people you kill can come back now if you decide to convert them and not just have to weirdly just. Put back on their sensors. Also, changed ferrochromic acid(Since that basically doesn't exist now) into politrinic acid for the deunomia recipe, so any observant command role can deconvert victims. Also did make the sting cooldown higher and the infection timer more generous(around 20 minutes for med to make and cure the patient).

## Technical details
YAML

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: The changeling infection sting now exists again! Also made deunomia more possible to get!


Xcybitt:
with 30 minute timers for convert and 10 minutes for a sting, lings getting more teammates is a factor but unlikely to happen(deunomia being able to fix infected people) and if they do, if you somehow ling every time 10 minutes is up, you on maximum can only convert like 12 people instead of what used to be the maximum(40). hopefully with these, it's gonna be better and more balanced. with the metashield, i'd assume command can and should tell the CMO to make deunomia given the situation. hope this helps

corresp0nd:
cling needs a better direction before this can be reenabled. it was not designed to be a conversion antag and simply making it into one ended up being more trouble than it was worth

PR #2032, opened 2025-11-19
Blob blocks Evac again. Minor Evac Blocking System Change.

ferynn:
## About the PR

Blob used to block evac for quite a while, but this was removed during a rework of various evac systems back a while ago. This was supposed to be temporary, so I did a little work and re-added it. This only happens post-blob discovery, and unlocks again after all blobs are defeated

I also made a slight modification to the system displaying evac blocking reasons, to support more than one antag blocking and trying to communicate a failure reason.

## Why / Balance

A discovered blob blocking evac significantly raised the tension of blob rounds. It made nuking the station feel like a big thing, something that actually meant something in character. Crew could still escape the blob through unofficial means, such as the cargo shuttle, but they couldn't take an NT approved way out. The captain had to weigh shirking their duty and leaving with the cargo shuttle, or ensuring the nuke went off. 

I think it added a lot to the creeping dread of the round. It's not uncommon for evac to be called midway through a blob round, before things got really tense. It also means that a blob that appears near the end of the round or evac cycle actually gets to have the crew engage with it.

## Technical details

Any blob in non-Default state now blocks evac. If evac was in progress when discovered, evac is cancelled. I also added a check in the ticker for blob to recall if it is in one of these advanced states, to account for recalls that happen without the console. However, I added a safety valve, where if this auto-recall does it three times, it will cease to recall the shuttle. While I don't think there are any situations right now that would cause that, this is to prevent any situations like the only Blob-Zombie combo from infinitely calling and recalling the shuttle. Rounds should always end, even if some loop akin to this starts.

I changed the string variable Reason in a CommunicationConsoleCallShuttleAttemptEvent to a List of strings. Now, if multiple antags subscribe to this event, such as nukies, the reasons are concatenated into a larger message, so that all reasons are displayed.

The blob evac blocking mechanism also checks that the blob in question is on the station. If it isn't for whatever reason, evac shouldn't be blocked.

## Media

Screenshot 2025-11-18 194546  

Blob blocking an existing evac call.

Screenshot 2025-11-18 195244  

Attempting to call evac with a discovered blob

Screenshot 2025-11-18 200143  

Attempting to recall with an active blob and nukies

## Requirements

## Breaking changes

Any reason for blocking evac in use of a CommunicationConsoleCallShuttleAttemptEvent  event should not be added to a list of strings, instead of by direct string assignment.

**Changelog**

:cl:
- add: A discovered blob blocks evac again.


TheHolyAegis:
And looking at the code, the blob will block evac before it reaches the first announcement before the sensors have detected the biohazard, which doesn't make sense

taydeo:
test this w the zombies gamemode, it broke evac w zombies and blob running at the same time

PR #2036, opened 2025-11-19
Prevent SM from ashing blob observer

Drywink:
## About the PR
The SM should no longer ash a blob player on contact.

## Why / Balance
The SM could kill the player observer on contact. The core would still be alive, but no longer controlled by a player.

## Technical details
Simply added the SupermatterImmune to the MobObserverBlob

## Requirements

## Breaking changes
@mycobiota had a fatal server error when attempting to add the SupermatterImmune to the blob observer, which I dont have, so I'm wary of that.

**Changelog**
:cl:
- fix: The SM should no longer ash any blob observer!


Drywink:
> Does a blob touching the supermatter cause the game to enter a broken state? I.e, the round can't end?

It doesnt do that fortunately. What it does is kill the blob observer, and the tile that touched it. The rest of the blob remains intact, but without a player controlling it.

mycobiota:
wonderful, thank you. i wonder why you didnt have the same crashing issue... did you test adding SM immune to the blob observer before adding the function to the SM code to prevent duplicate SM immune components?

PR #2037, opened 2025-11-19
Science can finally learn how to make their own Implants

A-Loose-Goose:
## About the PR
Adds a new Tier 2 Civilian research to science, "Homebrewed Implants". This simply unlocks the Bike Horn, Light, and Sad Trombone implants to the medical Techfab and Protolathes, as well as the Mindshield and tracking implant to the security Techfab. Mindshields and Tracking implants are available in Emagged Protolathes but that's mostly a 'I did it because I can' rather than it has any actual use case scenario. The only way I could see this being handy is a friendly thief during a Revolution or something.

## Why / Balance
Besides fluff for the fun implants, the only way to gain any implants is to buy them. Which I find boring. I feel like having alternate ways to get Implants rather than pray Cargo is financially stable adds more ways to tackle conversion antags. Plus since the Fun Implants would actually be accessible now, maybe we can actually see them being used now.

## Technical details
YAML. Just added new research and recipes to lathes really. Besides that, one FTL.

## Media
Media No 1
(Aware this shows as Tier 3, I did test, it is now Tier 2, I'm just too lazy to get another picture lol)
Media No 2
Media No 3

## Requirements

**Changelog**
:cl: ALooseGoose
- add: Science can now research most NT Branded Implants through Civilian Research!


taydeo:
make this level 2

A-Loose-Goose:
Done, will probably need testing since I dunno how balanced this is in all honesty but yeah

A-Loose-Goose:
Why do i bother…

PR #2048, opened 2025-11-20
Large amounts of Rev conversions can force open revolt. Also Gamma Alert.

ferynn:
## About the PR

If the Head Revs convert enough people to be larger than a given fractional threshold of the total station population, open revolt will be forced to be called. Certain types of conversions count for more 'effective' people, to show the difference between converting a random Assistant versus the Captain. Once the threshold specified is past, open revolt is forced, and an enforced Gamma alert will be called. If open revolt was called before this point, bu the threshold is still passed, Gamma alert will also be passed.

## Why / Balance

There are a few different reasons behind this. Sometimes HRevs take the logical approach of spending as much time as possible growing their revolution, even past the point where they could likely win, often waiting until the full two hours passes to engage the revolt. This is I think a decent bit of Rev related frustration, as even a good HRev is incentivized to delay their revolt as long as possible.

This changes that by forcing them to act after a certain density of conversions.

Another reason for this, namely the Gamma alert aspect, is to give a better delineation on the state of the station during a revolt, and the expected level of "Extraordinary measures" required. In cases before Gamma is called, the normal station is deemed to still be in normal order. Meaning procedures need to be followed, generally Sec should be the ones expected to deal with the revolution, and the Station is generally assumed to be in a crisis, but not like a total breakdown of standard procedure.

After a certain density of conversions though, this assumption breaks down in practical gameplay. When huge swathes of the station are converted, sometimes even Command or Sec, it forces non-traditional roles to respond, and sometimes in ways that break normal good practice. 

The Gamma alert gives a good signpost for this. Gamma alert is a step beyond Red, and is when Martial Law is enacted and SOP is suspended. It's not just something that means ERT, but a useful alert to describe the expected response in situations like this, where the station is tearing apart in every way.

## Technical details

Each conversion or deconversion changes a RevHeat value by an amount related to the converted role. 1 unit of RevHeat is equivalent to 1 "normal" member of crew.

| Role| Heat Change | Comments 
| ------------- | ------------- |------------- | 
| Magi, NTR|  20  | By converting them instead of killing them, you did a personal insult to NT| 
| Captain, HOS  |  10  | The centerpiece of Rev responses  |
| Command, IAA  |  6  |  Doesn't include above roles | 
| Security |  3  |   | 
| General Crew  |  1  |   | 

The total RevHeat value is compared to a threshold determined by the number of Minds with crew jobs on the station multiplied by a fractional multiplier.

Currently this multiplier is .6, so 60 percent, but could be changed to really anything.

Right now, beating the threshold without Open Revolt being called, instantly calls it, with the typical delay before announcement. I did plan originally on giving the HRevs a window to call it themselves before that, but I didn't get around to it, so it's like this lol.

## Media

Screenshot 2025-11-19 180632

An assistant being converted, and incrementing the RevHeat by 1.

Screenshot 2025-11-19 180654

Next they are deconverted, and the value is reduced back to 0.

Screenshot 2025-11-19 181042

An example of a security officer being converted, and the RevHeat being incremented by 3 instead.

https://github.com/user-attachments/assets/7b6911d5-f093-4f95-894c-5e15341101cf

A clip of a HRev exceeding the threshold, forcing Open Revolt, and calling a Gamma alert.

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Getting enough rebel converts can force Open Revolt.

### DIFF
--- a/Content.Server/GameTicking/Rules/RevolutionaryRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/RevolutionaryRuleSystem.cs
@@ -21,6 +21,7 @@
 // SPDX-FileCopyrightText: 2025 Tadeo <td12233a@gmail.com>
 // SPDX-FileCopyrightText: 2025 Tay <td12233a@gmail.com>
 // SPDX-FileCopyrightText: 2025 ferynn <witchy.girl.me@gmail.com>
+// SPDX-FileCopyrightText: 2025 rottenheadphones <juaelwe@outlook.com>
 // SPDX-FileCopyrightText: 2025 slarticodefast <161409025+slarticodefast@users.noreply.github.com>
 // SPDX-FileCopyrightText: 2025 taydeo <td12233a@gmail.com>
 //


88tv:
Magi/NTR being worth 20 crew members for this is hilarious. If HRev players see this, though, they may artificially avoid converting centcomm employees even if the opportunity is presented to them. Something to consider...

PR #2049, opened 2025-11-20
Armor Stun Duration Decrease

Xcybitt:
## About the PR
I think more people should get gunned down in the streets. Armor should have more buffs against stuns other than the stun res. With this, stun damage would be more a disarm option, which I feel would make it better feeling for the person being stun meta'd. Also, getting shot in the streets is what most antags want anywho, which this would help with. You CAN still arrest someone, but you need more equipment to do so(Flash, bola, and tear gas). I hope with this, security can use their lethals more against non petty criminals, as at the moment nonlethals are kinda meta. 
Also this doesn't even work atm, the code's fucked. This is moreso prep for when it in the future gets fixed.
Also maybe need to buff the void cultist, they may just be for ganking civillians if this change is merged.

## Why / Balance
People really don't like stun meta. With this, armors would make you be able to move 1 second before a cuff gets done on you if you got stunned RIGHT in front of a seccie and they cuff you immediately OR just knock you down since your armor negates stuns. With this, I hope security can gun down antagonists instead of just cuffing them.

## Technical details
pure unbridled YAML

## Media
(THIS DOESN'T WORK YET)

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Armors now have stun duration reduction, but it doesn't work at the moment.

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/armor.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/armor.yml
@@ -37,6 +37,7 @@
 # SPDX-FileCopyrightText: 2024 Джексон Миссиссиппи <tripwiregamer@gmail.com>
 # SPDX-FileCopyrightText: 2025 QueerCats <jansencheng3@gmail.com>
 # SPDX-FileCopyrightText: 2025 Tay <td12233a@gmail.com>
+# SPDX-FileCopyrightText: 2025 Xcybitt <197952719+Xcybitt@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2025 Xcybitt <xcybitt@gmail.com>
 # SPDX-FileCopyrightText: 2025 ferynn <117872973+ferynn@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2025 ferynn <witchy.girl.me@gmail.com>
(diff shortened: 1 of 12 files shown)


ferynn:
I know this is a work in progress, and as you said it doesn't work yet, but it looks like you're intending on making many armors give full stun immunity? Is that what you meant by putting the 0.0 modifier in?

Just trying to get a feel for your intentions here.

mikestrongarm1:
does this buff ALL armors or just antagonist's armors? if it's all of them this seems problematic because often times you can find armor vests in maints and not all people finding armor in maints are antags. if i wanted to then take that vest from them and they refuse this makes it much harder to peacefully arrest them. I feel like in order to nerf stuns armors should simply be given a lot of stun resistance. also, like 88tv said this kinda seems like it would incentivize sec to go lethal more often which is bad for roleplay between the antag and security. so i also think this change is unnecessary.

Xcybitt:
Probs a bad change come to think of it, maybe if I think of a better way to make it, but for now probs closed, thanks for the feedback

Xcybitt:
honestly, how even to start fixing stun meta in a way that makes nonlethals still possible while not being the go-to for actual gunfighting combat is... a bit. might just need more experience and time to get an inkling of an idea how to

PR #2076, opened 2025-11-22
Nukie Mail

Xcybitt:
## About the PR
Nukies can spawn in mail now. Basically singular grand lottery nukies.

## Why / Balance
It'd be REALLY funny. Also more strife in a round if there's none, I think the idea's fun but I'd understand since it's a bit much. It's REALLY rare so maybe it can work?

## Technical details
YAML

## Media

https://github.com/user-attachments/assets/172fd1d3-7947-46d8-8734-bcb9f01c63b4

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Added nukies to mail!


SigmaTheDragon:
> i... cannot think of any possible reason why this would be considered a good idea this is a thing that would be funny only like, once or twice, people already hate getting EMP mail and this is that but 10x worse
> 
> not to mention the implications of hard syndie contraband (this case the blood red suit c20r and viper) just being something that can apear from mail mail having a small chance to just fucking kill you does not sound fun for the person on the recieving end

To be fair the weight of it is 0.005 (the weight of getting syndie contra like the toolbox is 0.1)

Xcybitt:
honestly this is just a pr i really needed to get out not for actual good content but for the idea alone

Idem778:
This will be worth it just for the clips we’re gonna get from it

PR #2082, opened 2025-11-23
Buffed Stabilized Stimulants

Xcybitt:
## About the PR
Added no slowdown from wounds on use for stabilized stimulants.

## Why / Balance
Why does stabilized stims and ONLY stab stims don't have the no slowdown comp.

## Technical details
YAML

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Made Stabilized Stims give adrenaline.

### DIFF
--- a/Resources/Prototypes/_Funkystation/Reagents/exotic.yml
+++ b/Resources/Prototypes/_Funkystation/Reagents/exotic.yml
@@ -119,20 +119,19 @@
   physicalDesc: reagent-physical-desc-exhilarating
   flavor: medicine
   color: "#04ea9a"
-  worksOnTheDead: true
   metabolisms:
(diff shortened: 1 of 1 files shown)


corresp0nd (changes requested):
w/ no max heal, revivop will completely replace necrosol's niche so hard veto on worksOnDead. it also doesn't really need the buff, so it should evenly heal 5/dmg per group if you want to use the even heal groups instead of the regular ones. i dont remember what adrenaline does and i dont really think stabilised stimulants need it but whatever

Dragon232347:
What everyone else said. It absolutely makes no sense for a non cryo chem to work on the dead. Im fine with a small buff though.

PR #2089, opened 2025-11-24
millions must bleed

Mori428:
## About the PR

nerfed bleed healing on gauze and medicated sutures and blood regen for saline
## Why / Balance

I feel that bleeding in its current state is too easy to deal with for med and these changes will hopefully make bleeding at least a bit more of a issue
## Technical details

a few value changes
## Requirements

## Breaking changes

**Changelog**

:cl:

-tweak: nerfed bleed healing on gauze and suture and nerfed blood regen on saline

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Specific/Medical/healing.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/Medical/healing.yml
@@ -28,7 +28,6 @@
 # SPDX-FileCopyrightText: 2023 liltenhead <104418166+liltenhead@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2023 metalgearsloth <comedian_vs_clown@hotmail.com>
 # SPDX-FileCopyrightText: 2023 vanx <61917534+Vaaankas@users.noreply.github.com>
-# SPDX-FileCopyrightText: 2024 Ilya246 <57039557+Ilya246@users.noreply.github.com>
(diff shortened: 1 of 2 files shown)


Dragon232347:
Bleed is THE most deadly damage in the game. I feel confident in saying 90% of combat has been ended by bleed damage. Ways to heal it do not need a nerf

mikestrongarm1:
while I agree that bleeding damage needs a buff this seems kinda extreme maybe set saline to 3 instead of 0.4 because if not at that point I think it's just easier to use blood packs on people instead of saline

mikethememer:
Bleed is extremely powerful and if you don't heal it you are DYING 
Bleed doesn't need a buff

Mori428:
to go in depth with why I think bleeding should be more dangerous I generally feel like bleeding is less dangerous than it should be. for medical bleeding is practically a non issue and can be easily dealt with by a singular gauze and a syringe of salty water. and for combat it is dangerous but just some gauze or a medipen can counter it. I think by making saline less effective it will force people to prep more around bleed when it comes to combat like bringing stuff like dex and making medipens and cauterizing more useful. I do fully think that saline does need to be nerfed but if it really is a issue I could look at adding a advanced chem that heals blood better. idk im gonna leave this open I do genuinely think these changes are good but if it get closed then ill think more about how I can make it work in a way people want.

corresp0nd:
nerfs like this don't encourage people to play around a certain thing being "dangerous," they just encourage people to use more resources

PR #2106, opened 2025-11-26
Worse winter coats

Xcybitt:
## About the PR
0.1. WHY. Emergency EVA suits were 0.3. Scarves are 0.5. Winter boots are 0.5. It deserves to be worse. It's now 0.4.

## Why / Balance
Why was it better than an actual suit made for space 😭  This should make winter suits less good for IPCs to walk into space(You can like just wear more winter gear to combat this though, which I don't exactly hate, since it means you HAVE to kit out and not just. Walk into space. Other specieses can't walk into space unless they actually prep for it or only walk for a certain duration or sip cryox.

## Technical details
YAML

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Killed winter coats. Go wear more so you don't freeze from space.


ferynn:
Full agree with this PR. People shouldn’t be able to put on a jacket and walk casually around in -100C air, which they can right now. They need to be way weaker.

PR #2175, opened 2025-12-02
Added wordle to pda

rex1431ify:
## About the PR

Wordle can now be played in the pda
## Why / Balance

Added because the pda has no games also tay asked me to add this

## Media

https://github.com/user-attachments/assets/a0bfcf30-4b05-45d9-b601-77a1ae959be2
## Requirements

## Breaking changes

**Changelog**

Add: new pda app wordle


mkanke-real:
Holy shit its real

whythelettuce:
I saw this PR and I absolutely love the idea of Wordle on PDA's- however I *am* a little concerned about the valid word list in WordleCartridgeComponent.cs? The comment for the valid word list states "List of valid 5-letter words for the game," but many words in the list are more than five letters (see: DIPTYCH, DOPIEST, BORDER), and it seems to stop giving valid words and start cycling through things alphabetically around the word "DOWRF." You'll see it start to just go alphabetically with DOWR**G**, DOWR**H**, DOWR**I**, etc. There are also no words that start with any letter after D? It just stops. I'm no expert, but it seems like something is off and I wanted to make sure this was known!

whythelettuce:
After diving a little deeper, it also seems that the code doesn't actually check to see if your guess is a valid word before allowing you to submit it? So theoretically, you could put any random five letters (QTYUI or BHJSP) and it would process it like any other word, using up a guess for a word that doesn't count in the list. It also seems to default to the word "WORDLE" before picking which word from the valid word list to use, which... is also six letters (WordleCartridgeComponent.cs, lines 12-15). May want to double-check that the word list is valid + you can't just type any random five letters in as a guess

PR #2190, opened 2025-12-03
HoS gets a N1985

Xcybitt:
## About the PR
HoS with the big magnum. What will he do. But genuinely, this PR makes a N1985(N1984 with half the rate of fire BUT a fun switch(full auto)), and puts it in the HoS' pocket roundstart, changing the MK58. I think the N1984 is really neat, and wish for it to exist somewhat in game. So here we are.

## Why / Balance
Realistically this is a worse WT550(Also like kill that thing, give the HoS a really fancy prototype pistol instead)/better det revolver. It's also just cool, and would distinguish the HoS more from other officers.

## Technical details
YAML

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Added the N1985, a N1984 with half the firerate but a full auto selector.
- tweak: Changed the MK58 the HoS starts with with a N1985, and also changed their backpack mag to the magnum one.


Xcybitt:
forgot that we're moving away from gunsec so

PR #2203, opened 2025-12-04
Cocoon balancing, and new bite ability for arachnids and giant spiders

Drywink:
## About the PR
Cocooning a target is now faster if they are incapacitated (critical or dead), sleeping, or stunned.

A new bite ability has been added for both the Arachnid specie and giant spiders (Shiva and Tarantula). This ability delivers an instant bite that injects 10 units of chloral hydrate into the target and notifies them of the action. Functionally, it works similarly to a changeling sting, but it is not silent and always alerts the victim.

This ability can synergize with the reduced cocooning time on sleeping targets, but due to the nature of chloral hydrate (putting targets to sleep after a random delay of 5–60 seconds and only for 2–5 seconds) combined with the fact that the target is informed, it remains difficult to pull off reliably.

## Why / Balance
This feature is experimental and may require balance adjustments.
I’m very open to suggestions or balance concerns to help refine this feature.

Reducing the time to wrap someone when they are incapacitated made sense, similarly to the handcuff.

The new bite ability provides spiders with a (theoretically) non-lethal way to capture someone alive, though its usefulness is intentionally limited by the short and random sleep duration and the warning given to the target.
For example, Shiva could use it as a non-lethal tool to capture criminals.

## Technical details
Added a new, fully customizable Bite ability. Its injected chemicals, cooldown, and description can all be configured for use in different systems.

## Requirements

## Breaking changes
Since it is in parallel to the cocoon fix PR I made, it might create some conflicts.

**Changelog**
:cl:
- add: New bite ability for arachnids and giant spiders.
- tweak: Reduce cocoon doafter time on incapacitated target.

### DIFF
--- a/Content.Server/Species/Systems/CocoonSystem.cs
+++ b/Content.Server/Species/Systems/CocoonSystem.cs
@@ -1,3 +1,4 @@
+// SPDX-FileCopyrightText: 2025 Drywink <43855731+Drywink@users.noreply.github.com>
 // SPDX-FileCopyrightText: 2025 Drywink <hugogrethen@gmail.com>
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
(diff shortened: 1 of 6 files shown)


TrixxedHeart:
There should _really_ a doAfter for the bite, can immediately see that being a huge issue.

Drywink:
> There should _really_ a doAfter for the bite, can immediately see that being a huge issue.

If this becomes an issue, your solution could work too, something like the cosmic cult stun speed. Ill make a fix with your idea if this ever become an issue.

PR #2256, opened 2025-12-09
Heart Surgery Doesn't Cure Rot

ferynn:
## About the PR

I removed the interaction where inserting a heart via surgery removes all rot from a body. Now, the only way to feasibly remove rot is through the chemical that does it.

## Why / Balance

I think this interaction was both nonsensical and it utterly trivialized rotting and death in general. It made no sense at all, from both a realistic perspective and a stylistic one, that putting a random heart in a rotten body would restore it. It makes rot something that is easily cured through a trivial surgery, as completely changing a person's heart is both easy and risk free.

There was no restriction on who's heart it was either. I'm pretty sure it was possible to take someone heart out, and immediately put it back in to cure their rot. 

It was a silly interaction, but not the fun kind of silly we want.

Removing this will make death scarier, and the risk of round removal higher. But it will also make people fear getting into those situations where death could happen and respect them. Rotting can still be reversed through Opporozidone with greater difficulty. 

## Technical details

Removed rot removal from insert heart surgery step prototype

## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: Inserting a heart no longer removes rot from a body.

### DIFF
--- a/Resources/Changelog/Funkylog.yml
+++ b/Resources/Changelog/Funkylog.yml
@@ -6588,3 +6588,9 @@ Entries:
         longer be worn on the belt.
   id: 770
   time: '2025-12-14T23:36:47.0000000+00:00'
+- author: vivry
+  changes:
+    - type: Tweak
+      message: Accent tooltips in the character creation menu now demo the accents
+  id: 771
+  time: '2025-12-15T01:16:26.0000000+00:00'
--- a/Resources/Locale/en-US/_Funkystation/traits/traits.ftl
+++ b/Resources/Locale/en-US/_Funkystation/traits/traits.ftl
@@ -7,22 +7,22 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later AND MIT
 
 trait-funky-name = Funky Accent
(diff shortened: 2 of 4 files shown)


Idem778:
So, I agree with this PR but just for information, the heart saves the rot level so if you put a heart from a rotted body into they’ll still be rotted

mikethememer:
I love being Rr because paramedic didn't check crew monitor for a few minutes!
Please don't, it's really annoying and rotting doesn't make death more meaningful, it only makes it that there is a timer for a single job on station to save you, which can be antag which can be not present etc

ferynn:
Rotting is not a fast process. It's not something that requires the paramedic to be glued to the monitor to avoid. It takes like 15 minutes. 

You also don't have to be found, recovered, and cured within that window either. Medical has plenty of ways to halt the progression of rot, so if they are swamped with cases, they don't have to worry about the bodies rotting. That is why body bags and the morgue exist. To hold bodies that they aren't actively treating, so they don't rot. It's not meant to just be a place for DNR'd bodies to be stored.

PR #2260, opened 2025-12-10
REMOVE SCIENCE

BrightNibbleston:
## About the PR

removed the flat pack from the research tree

## Why / Balance

I removed it because Its QOL slop and reduces interdepartmental interactions. It cuts engineering out of the equation completely because everyone and their mother has a multitool with present day power creep. (thanks joules for helping write the why part)
in my own words its because it makes science not need to go and talk to engineering and gives engineering even less content than they are already starved for.

## Technical details

removed the flatpacks research thing in industrial.yml

## Requirements

**Changelog**

:cl:
- remove: Removed the flat pack from research


BrightNibbleston:
> Instead of removing it, consider locking it behind higher tiers of research.

ngl, i really dont think even that would really make it balanced, because than it just makes that the research you want first and makes every other branch less important a lot more, so people get even less civ and engi stuff from sci, along with the fact its still op as you can remove any building, or storage limitations that one might have otherwise

PR #2289, opened 2025-12-13
Forever Alcohol

PurpleTranStar:
## About the PR
This PR Adds Forever Alcohol, a unique type of alcohol that persists in the bloodstream until ethylredox is administered.

## Why / Balance
We already have forever weed, why not forever alcohol? Now you can suffer twice as hard!
Forever alcohol also doesn't heal dwarves, so that's fun.
PLEASE ADD AN OPTION TO DISABLE SCREENSHAKING THROUGH THE ACCESSIBILITY BEFORE MERGING THIS
Or don't.
I won't tell.

## Requirements

**Changelog**

:cl:
- add: Added forever alcohol, a special alcohol made from Cogchamp, Long Island Iced Tea, Pan Galactic Gargle Blaster, Ethanol, and Liquid Exotium


corresp0nd:
i cant merge something like this in good faith if we know that it will cause accessibility issues

PR #2315, opened 2025-12-15
Remove hit chance on prone targets

Drywink:
## About the PR
Removed the chance to hit a prone target by accident.

## Why / Balance
The chance for bullets to hit prone targets was originally there to compensate for how hard it was to aim at someone on the ground. With bullet prediction now in place, aiming is much more accurate, so this extra hit chance isn’t really needed anymore.

Keeping it caused some annoying side effects: dead bodies would keep getting shot, making revives harder, and piles of corpses could be used to block bullets.

## Requirements

**Changelog**
:cl:
- tweak: Bullets no longer have a chance to hit downed target by accident.


ferynn:
The hit rate thing is also bugged, and just auto hits prone things even if it doesn’t look like it does. I put in a reversion PR for this a while back because of that.

Removing the prone hit chance thing would be better than nothing. Could let us add more nerfs to prone combat as needed on top of it.

Drywink:
One solution for that specific case you mentioned could also be to reveal the stealth suit user if they attack.

PR #2337, opened 2025-12-17
Zappy Insuls

YaraaraY:
## About the PR
This was such a funny discussion on Discord

This PR removes MV and HV protection from regular insuls and combat gloves, meaning you get zappy zapped when trying to snip HV and MV wires. Grilles and doors are not affected; any level on insulation still protect you from them, regardless of what type of cable is running under them. It adds three new pairs of insuls; CE's insuls and Arc Gloves, available in the Engi-vend. These both provide full protection from all voltages. And, the third pair; blood-red combat gloves. These appear where previous Syndicate-related combat gloves did, so Nukies spawn with these for instance, and the suspicious toolbox has a pair. They are NOT super insulated. Go mug the CE.

Oh, I also gave Engineers material bags in their lockers because someone suggested it to me while I was messing with the fills. Uhm. Oops.

This also edits several .YMLs to add the full insulation to some gloves that I felt deserved them.

## Why / Balance
Now when dealing with live MV and HV cables, engineering will need to go and disable the relevant substation; potentially cutting off power from nearby departments in doing so. This adds inter-departmental conflict, leads to engis spending more time in maints, and adds a lot more overall interaction to the round than simply going over and placing an MV cable. 

When in situations where they need to cut MV or HV cables, engineers can either ask their CE, grab a pair of arc gloves, or use their RCD. This gives a new niche use for the RCD, however small it may be.

This changes how sabotage is done, as instead of snipping one wire, you'd instead break into the substation room and disable it. This is more fun for Engineering to repair, and affects MANY APCs, rather than whichever the MV would.

## Technical details
YAML maxxing, new sprites.

## Media
image
image
image

## Requirements

**Changelog**
:cl: AraiMaia
- tweak: Insulated Gloves no longer protect you from MV and HV cables; try the Substation instead.
- tweak: Engineers and Atmos now get material bags in their lockers, if they'd like to sacrifice their toolbelt in exchange for carrying lots of materials.
- add: CE gets their own pair of gloves with full insulation.
- add: Added Arc Gloves, available in the Engi-vend, are gloves with full insulation. There's just one, though. Share!
- add: Added Blood-red Combat Gloves, new gloves with insulation, replacing the regular combat gloves in Syndicate related loadouts and uplink. They're not Super Insulated though.

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/Hands/gloves.yml
+++ b/Resources/Prototypes/Entities/Clothing/Hands/gloves.yml
@@ -374,8 +374,6 @@
   - type: Clothing
     sprite: Clothing/Hands/Gloves/mercbattle.rsi
   - type: Insulated
-    coefficient: 0.0
-    maxSiemensThreshold: 4
   - type: Butcherable
     butcheringType: Knife
     spawned:


ferynn:
I know your intent is for power related projects leading to engineering shutting off substations, but I think it would be much more common for them to just use the RCD, or force whoever has immune gloves to fix it.

I can’t really see engineers ever really shutting off substations when there are this many ways to bypass the risk.

TheSecondLord:
I don't hate this overall but I think this will be bad for nukie rounds. It will increase the number of deaths caused by bad luck and is disproportionately punishing for newer nukies/nukies who haven't read the entire changelog and aren't aware of this fork-specific change. You mention it making sabotage more interesting, but it entirely removes the opportunity for certain stealth activities as an antag, such as breaking someone else/yourself out of perma or breaking into the bridge/captains office from space, since both pretty much always have windows with HV grilles. I think if you're wear the super valid evil gloves you should be allowed your HV protection.

PR #2339, opened 2025-12-17
stun nerfs

Mori428:
## About the PR

lowered the stun damage on stun baton and prod to 35 and disabler to 25, and added stun resist to web vest and acolyte armor
## Why / Balance

this change shouldnt effect stunning unarmored enemies but will make stunning anyone with stamina resistance take more hits and I avoided nerfing beanbags because I think they are fine and smgs because I didnt wanna think very much about them and they both require research so its less likely sec will go for them
## Technical details

a few number changes and a few lines of code added for stamina resistance
## Requirements

## Breaking changes

**Changelog**

🆑 
- tweak: nerfed disabler, stun baton and stun prod against stamina resist
- tweak: web vest and acolyte armor now have stun resist

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/armor.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/armor.yml
@@ -35,6 +35,7 @@
 # SPDX-FileCopyrightText: 2024 metalgearsloth <31366439+metalgearsloth@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2024 voidnull000 <18663194+voidnull000@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2024 Джексон Миссиссиппи <tripwiregamer@gmail.com>
+# SPDX-FileCopyrightText: 2025 Mori <salvanson10@gmail.com>
 # SPDX-FileCopyrightText: 2025 QueerCats <jansencheng3@gmail.com>
 # SPDX-FileCopyrightText: 2025 Tay <td12233a@gmail.com>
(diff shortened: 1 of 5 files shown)


mikestrongarm1:
I don't disagree with stuns being nerfed however I am of the opinion that if stuns are to be nerfed that you should only give armors stun resistance rather than nerfing the actual stun weapons themselves

Mori428:
> Probably not a necessary change. Any distinct nerf to stuns inherently promotes lethals, which... Do you really want a dead antagonist instead of a cuffed one?
> 
> Otherwise, I fail to see the point in making the antagonist's objectives any easier than it currently is. Crew feels weak enough as it is, especially on low population numbers.

this change should only really be making stuns less effective against anyone armored which I think if someone has armor lethals should be a more effective option because I generally think its kinda lame when someone will get destroyed by non lethals when going loud it feels kinda underwhelming

Mori428:
> I don't disagree with stuns being nerfed however I am of the opinion that if stuns are to be nerfed that you should only give armors stun resistance rather than nerfing the actual stun weapons themselves

the stun weapons had weird damage break points that make how many hits it would take against stun resist kinda weird and this makes it more consistent I think like for instance despite the armor vest having 15% stun resist it still goes down after 3 stun baton and 4 disabler hits or the sec hardsuit that takes 4 baton hits but 4 disabler hits just because of the weird damage numbers armor can be buffed and if this change sucks I can just do that instead but I think this option makes sense

taydeo:
Changes for stuns are planned but this is not going about it the right way.

PR #2346, opened 2025-12-18
Update Medal Awarding To Be More Intuitive

TrixxedHeart:
## About the PR
Improves the code for medal awards to prevent accidental self awards due to how it works.

Original behavior:
Equip the medal onto the neck slot of someone for it to be awarded. This caused issues when players attempted to edit the award's reasoning, as equipping the medal took priority over editing the reason.

New Behavior:
The medal no longer awards upon being equipped to the neck slot. Instead the player must hold the medal in their hand and click onto the entity they wish to award it to. This gives a 2 second doafter, of which it will auto equip it to the awardee's neck slot if available. If it isn't, it will be put into the awardee's hand. If no hands are available it'll remain in the awarder's hand

This has the incredibly humorous side effect of allowing any entity to be awarded medals.

## Why / Balance
Prevents accidental:
image

## Technical details
Changed the behavior to use a doafter based on clicking an entity

## Media

https://github.com/user-attachments/assets/d4a7fb98-bfe3-4f9c-b254-2b175f990c62

image

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Medal awarding now works by clicking onto an entity with the medal in-hand instead of being placed on the neck slot. This should prevent accidentally awarding yourself a medal when you meant to edit the reasoning.
- add: You can now award medals to any entity. Great job, fuel tank!

### DIFF
--- a/Content.Server/_Impstation/CrewMedal/CrewMedalSystem.cs
+++ b/Content.Server/_Impstation/CrewMedal/CrewMedalSystem.cs
@@ -1,4 +1,6 @@
-// SPDX-FileCopyrightText: 2024 AirFryerBuyOneGetOneFree <jakoblondon01@gmail.com>
+// SPDX-FileCopyrightText: 2025 AirFryerBuyOneGetOneFree <jakoblondon01@gmail.com>
+// SPDX-FileCopyrightText: 2025 Mora <46364955+TrixxedHeart@users.noreply.github.com>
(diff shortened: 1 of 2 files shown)


TrixxedHeart:
> Why wasnt this added earlier… I wanted to do this to Greg but I couldnt

Because it was initially cherry picked from Imp without any extra changes besides to fix a minor issue

Xcybitt:
i can't wait to give the clown's banana peel a medal and the HoS' baton a medal

PR #2366, opened 2025-12-21
Welding Sparks Port

YaraaraY:
## About the PR
Ports [this PR from GoobStation](https://github.com/Goob-Station/Goob-Station/pull/5476), adding a cute little animation to welding.

## Why / Balance
CUTE!!!

## Technical details
Changed namespaces to fit our codebase, but otherwise its a straight port.

## Media

https://github.com/user-attachments/assets/89452470-d075-4cc7-b2bc-7c3fe16dc4ac

https://github.com/user-attachments/assets/72a865d6-af73-4ee8-bcb9-a79ee46feaf0

## Requirements

**Changelog**
:cl: SabreML, AraiMaia
- add: Added a cute little animation to welding doors and objects, ported from GoobStation, made by SabreML (GitHub)

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Doors/Airlocks/external.yml
+++ b/Resources/Prototypes/Entities/Structures/Doors/Airlocks/external.yml
@@ -43,7 +43,7 @@
   - type: Wires
     layoutId: AirlockExternal
   - type: WeldingSparksAnimation # Goobstation
-      startingOffset: -0.5, 0
+    startingOffset: -0.5, 0
 
 - type: entity
   parent: AirlockExternal


YaraaraY:
> My fault in the original PR! `OnSpawnedWeldingSparks()` uses `GetEntity()` for the target and the spark entities, but it looks like it only actually verifies that `targetEnt` exists, with `TryComp()`. `sparksEnt` just gets passed along to the animation system whether it's valid or not.
> 
> The fix should be just changing it to `TryGetEntity(ev.SparksEnt)` and adding a check. I'll double check that then post a fix.
> 
> (I have no idea _how_ it could possibly be called without the spark entity existing, but it's all client-side code so anything can happen there I guess.)

evil evil clientside code... thanks for the headsup! will port over your fix when i have the time <3

PR #2393, opened 2025-12-23
Minor additions to Security SOP and RoC

jackel234:
## About the PR
I added a line to the Security Officer SOP for Code Green, and a line to the RoC for Security.
These being:
[bold]17.[/bold] Security Officers must wear a uniform that makes them identifiable as a member of the department.
[bold]9.[/bold] Security Officers are discouraged from entering maintenance without a valid reason. If they must enter maintenance for one reason or another, it is highly recommended that they do so with a partner.

## Why / Balance
The SOP had mentioned that secoffs must wear a uniform in accordance with the RoC, but the RoC never really defined what said uniform could be. This addition, while vague, at least provides a solid guideline to work off of. Secondly, it was already treated as a phantom rule that security could not enter maints on green, yet nowhere in Space Law or the SOP was it actually mentioned, just suggested within the Security Officer section of the guidebook. Ultimately, it'd be best to codify this for future clarity.

## Technical details
I changed the xml files in vim to add the new lines :)

## Requirements

**Changelog**
:cl: amethystowo
- tweak: Added a new line to the Security Officer SOP, codifying that they should not be in maints on Green. Also added a line in their RoC that loosely defines what their uniform should be.

### DIFF
--- a/Resources/ServerInfo/Funkystation/Guidebook/Security/RulesOfConduct.xml
+++ b/Resources/ServerInfo/Funkystation/Guidebook/Security/RulesOfConduct.xml
@@ -32,7 +32,7 @@ SPDX-License-Identifier: MIT
 
   [bold]11.[/bold] Security Officers shall not be absent from duty assignment.
 
-  [bold]12.[/bold] Security Officers shall not be critical or talk negatively about other Nanotrasen security officers or any member of Command while on-duty.
+  [bold]12.[/bold] Security Officers shall not be critical or talk negatively about other Nanotrasen Security Officers or any member of Command while on-duty.
 
   [bold]13.[/bold] Security Officers shall not join a union or entertain the idea of a union.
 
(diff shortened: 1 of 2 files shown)


jackel234:
it's still a good addition regardless, as it's something most HoSes already bar their officers from doing, so codifying it makes sense. it's a good rule to have either way, as it keeps security out of maintenance when antagonists are at their weakest in a meta sense, and makes sense from an ic perspective as well

88tv:
> The SOP had mentioned that secoffs must wear a uniform in accordance with the RoC, but the RoC never really defined what said uniform could be. 

So just add it to the RoC?

> it was already treated as a phantom rule that security could not enter maints on green

Nope. If secoffs want to go get stunned by the changeling in maintenance, let them do it. As much as it's not really thematic for security officers to go into maintenance, I prefer this being enforced by your **boss** rather than **corporate code**.

PR #2452, opened 2025-12-28
kill disabler smg

taydeo:
not really fun + we want to move away from stuns like these in general

:cl:
- remove: Disabler SMGs are no longer obtainable.


taydeo:
mald self merge i got stunned by one as a nukie

JohnnyDangerSpace2:
wasd to move your character. hope this helps!

PR #2462, opened 2025-12-28
Disabler Nerf(DO NOT MERGE UNTIL PROJECTILES CAN INJECT INTO SPESSMEN AGAIN)

Xcybitt:
## About the PR
After discussion, made disablers slow people instead of damaging their stamina. Thank fuck everyone parented off the disabler bolt. Added a "disabler mix" chemical that's just in the disabler projectile. Also uh made disablers nonlethal as shit(No more ion), to give it the capability to be used by thieves.

## Why / Balance
As someone put it, low risk, high reward. Having to melee to actually get someone horizontal would make for chases, more fun, etcetera. Also, more nonlethal options would be on the table to get someone arrested.

## Technical details
YAML

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Disablers now no longer stun people, but slow them down given enough shots on target.
- tweak: Made disablers be able to be used by thieves.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -75,9 +75,9 @@
 # SPDX-FileCopyrightText: 2025 pa.pecherskij <pa.pecherskij@interfax.ru>
 # SPDX-FileCopyrightText: 2025 slarticodefast <161409025+slarticodefast@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2025 taydeo <td12233a@gmail.com>
-# SPDX-FileCopyrightText: 2026 Xcybitt <197952719+Xcybitt@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2026 ThatOneMoon <91613003+ThatOneMoon@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2026 ThatOneMoon <juozas.dringelis@gmail.com>
+# SPDX-FileCopyrightText: 2026 Xcybitt <197952719+Xcybitt@users.noreply.github.com>
 #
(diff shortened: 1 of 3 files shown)


ferynn:
Depending on how fast it takes to come into effect, and how large the slow, this could be stronger than base disablers in some ways lol. Worse against like nukies, as you can’t take them down, but for general use this now slows on 1 hit instead of 2 or 3.

Slow downs are killer, especially when you’re trying to run away. This looks like it ranges from 10 percent slow down to 25 percent slowdown. How many disabler shots does it require to hit these accumulation thresholds?

Xcybitt:
> Depending on how fast it takes to come into effect, and how large the slow, this could be stronger than base disablers in some ways lol. Worse against like nukies, as you can’t take them down, but for general use this now slows on 1 hit instead of 2 or 3.
> 
> Slow downs are killer, especially when you’re trying to run away. This looks like it ranges from 10 percent slow down to 25 percent slowdown. How many disabler shots does it require to hit these accumulation thresholds?

a hit is 5u, assume we take a second each so 3u/s, 5 second to first threshold of 10 slowdown, then  4 seconds more for it to hit 25 slowdown

taydeo:
this dont even work yet

Xcybitt:
> Ok, I'm going to be real, I hate this pr, like genuinely. To be clear, I don't hate the concept so much as I hate the implementation. This is the most hacky, yaml squire ass implementation of this concept you could have done. What happens when someone already has a mix of chems in their bloodstream, and this now has to share space with that? What would happen if someone took charcoal pills or pentetic acid and completely negated the effects of the disabler, in a completely unimmersive way. Like this has no thought put into the actual implementation, and just takes the easiest, hackiest way out, instead of implementing the idea properly.

that's fair, i just don't exactly have ideas for makin people slow that isn't chems or stamina, could theoretically make this a temp gun but that would still have it's own issues.

PR #2470, opened 2025-12-29
Changeling Tweaks

ferynn:
## About the PR

New set of ling changes and tweaks.

Fleshmend and Anatomic Panacea now consume biomass on usage. Fleshmend costs 8 biomass (out of 30 max). Anatomic Panacea costs 4 biomass. Their usage cooldown is lowered from 30 seconds to 10 seconds.

Regenerative Stasis has had its biomass cost increased from 1 -> 4 biomass. The chemical cost to wake up has been shifted from 60 to 50.

The negative visible effects of low biomass are at lower total biomass thresholds now. This means it it more difficult to reach those stages.

Fleshmend was recosted at 4 evo points instead of 5. Fake armblade has been changed to 4 evo points instead of 5. Augmented vision has been increased in price from 4 to 6 evo points. The changeling spacesuit has been recosted at 2 evo points from 4.

Changelings can no longer rot.

## Why / Balance

Biomass has been in a strange position since the ticking drain was removed. I want to add it as more of a consumable resource, one less easy to get than chemicals. It requires the consuming of corpses, and is more valuable. As lower biomass is now more of an internal resource to track, than a ticking social stealth clue to others, I made it more difficult to get the "visible" signs of low biomass. This will hopefully prevent normal usage from causing changelings to vomit everywhere.

A biomass cost was added to the two non-stasis healing abilities, as it prevents a ling from hiding and healing indefinitely, cost and risk free. Now, they must consume if they wish to heal. Fleshmend in particular is quite expensive. The base biomass cost of stasis was increased as well, as there is no longer an associated ticking cost.

Evo point wise, fleshmend was lowered due to the nerf, and to keep it more in line numerically with other abilities. Fake Armblade and Spacesuit abilities are pretty bad, so I'm making them slightly cheaper at least. Augmented eyesight is the strongest changeling ability, and I'm costing it to represent that.

Finally, changelings in stasis really should not rot at all. The perishable component has had their "time to rot" component maxed out, a duration longer than games last.

## Technical details

Mainly just yaml tweaks. 

A function was added to SharedRottingSystem to allow alternate base rotting times to be assigned. The RotAfter value for a changeling's PerishableComponent was set to TimeSpan.MaxValue with this. 

Biomass thresholds for negative effects have a different fraction now in the C# code.

## Media

Screenshot 2025-12-28 223729

Changed to cost biomass.

Screenshot 2025-12-28 223749

Biomass after using fleshmend once.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Fleshmend and Anatomic Panacea now cost biomass instead of chemicals. Cooldown reduced.
- tweak: Regenerative Stasis costs more biomass to use. Chemical cost to rise again is slightly lowered.
- tweak: Several changeling abilities had their evolution point cost changed.
- tweak: Changelings must lose more biomass to display visible signs to others.
- fix: Changelings can no longer rot

### DIFF
--- a/Content.Shared/Atmos/Rotting/SharedRottingSystem.cs
+++ b/Content.Shared/Atmos/Rotting/SharedRottingSystem.cs
@@ -3,6 +3,7 @@
 // SPDX-FileCopyrightText: 2024 Kara <lunarautomaton6@gmail.com>
 // SPDX-FileCopyrightText: 2024 Nemanja <98561806+EmoGarbage404@users.noreply.github.com>
 // SPDX-FileCopyrightText: 2024 Tadeo <td12233a@gmail.com>
+// SPDX-FileCopyrightText: 2025 ferynn <witchy.girl.me@gmail.com>
 // SPDX-FileCopyrightText: 2025 taydeo <td12233a@gmail.com>
 //
 // SPDX-License-Identifier: MIT


Dragon232347:
maybe im still malding but lings sre increidbky under powered. I dont think we should be increasing the cost of any of their abilities ngl. While i like the biomass drain bc it forces you to actually go and kill people this makes getting caught once even more of a death sentence than it already was.

ferynn:
> i'm slightly concerned that the lack of the perishable component might just make it easier for ppl to metagame lings

Yeah. I’m a little concerned too. I’m probably going to go back in, and just see if I can keep that component, and just make the time to rot gigantic

ferynn:
> i'm slightly concerned that the lack of the perishable component might just make it easier for ppl to metagame lings

The perishable component is now there, and the corpse looks normal. It just takes longer than any round would ever last to actually rot.

Jack-Silver:
> maybe im still malding but lings sre increidbky under powered. 

100%
lings are not that strong.
i actually think by doing this all you're gonna do is make it so lings hide in the shadows even more and do nothing at all if they're on low biomass

PR #2483, opened 2025-12-29
Experimental Changeling changes - Removing bloodtesting

miosar:
## About the PR

Removed the changeling bloodtest exploding. To identify a ling, you will now have to see them perform changeling activities, just like identifying a heretic due to them casting spells.
## Why / Balance

It is actively encouraged to not stay hidden - when a security officer or a command member randomly demanding a blood test cannot be denied without causing suspicion ( which is usually followed by a wordless stun + cuff if security has been briefed and changelings are known to be on station ), it is much safer to attempt to just kill as many people as possible with your hivemates.

These changes should hopefully increase the viability of not being caught as a changeling without forcing a pacifist playstyle.

There are, of course, potential issues that may arise with the more difficult identification of changelings. That may still however be possible due to the following reasons:

    -Changeling bodies not rotting
    -Changeling abilities are clearly visible, such as the armblade, EMP scream and other things clearly able to be recognized by command
    -In the case of two identical people, command may act the same way as they do with paradox clones - It is much more fun to play trivia than to wordlessly take blood from both people while they wait in the brig!

## Technical details

Removed the kudzu and explosion effect from the changeling blood reaction in the centrifuge.
## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Removed changeling bloodtesting. Find out who the impostor is the old fashioned way!

### DIFF
--- a/Resources/Prototypes/_Goobstation/Changeling/Reagents/biological.yml
+++ b/Resources/Prototypes/_Goobstation/Changeling/Reagents/biological.yml
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: 2024 Piras314 <p1r4s@proton.me>
 # SPDX-FileCopyrightText: 2024 Tadeo <td12233a@gmail.com>
 # SPDX-FileCopyrightText: 2024 username <113782077+whateverusername0@users.noreply.github.com>
+# SPDX-FileCopyrightText: 2025 Miosar <74148670+miosar@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2025 taydeo <td12233a@gmail.com>
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later AND MIT


Komonighub:
who would do this

Terkala:
Doesn't this mean changeling will just be outed by their blood being the wrong type? So moths/slimes/diona/vox changelings would all be spotted instantly from having the wrong bloodtype. And this makes it so security has no way to identify a changeling other than catching them in the act. So even if security knows someone is a changeling, they can't execute them, and will have to let them go since they can't prove it. All of the listed examples are things changelings can just choose not to do and wait out whatever inspection they're under.
 
 Overall I don't think this change is well thought out.

taydeo:
> Doesn't this mean changeling will just be outed by their blood being the wrong type? So moths/slimes/diona/vox changelings would all be spotted instantly from having the wrong bloodtype. And this makes it so security has no way to identify a changeling other than catching them in the act. So even if security knows someone is a changeling, they can't execute them, and will have to let them go since they can't prove it. All of the listed examples are things changelings can just choose not to do and wait out whatever inspection they're under.
> 
> Overall I don't think this change is well thought out.

bloodtesting is also a problem, and seeing that the more and more i lean to just outright removing the antag for now (since we have i believe 2 design doc prs on our doc repo) is healthier than whatever we can do for it right now. bloodtesting is a mechanic that i believe does a lot more damage than just removing it entirely. my reasoning for approving this is to encourage more mystery even if that means giving changeling an advantage, because yeah, the entire point of the antag is to be a super scary mystery killer.

also, on the point of RRs, while i dont really care that "innocent" people get RR'ed, i do care that the experience is novel and engaging. this is a problem with changeling, the ideas and concepts presented here are Cool but there is nothing really concrete in-game backing up the way it wants to be portrayed as. as for the points of security not being able to do anything about it, i think this warrants a greater discussion on how we want changeling to be played.

there is no avenue that a Super Killer Maints Monster is going to be fair in terms of "You did this, so you get X punishment." so we should avoid that kind of thinking.

it really is just a matter of picking our poison if we so desperately want changeling in the game still, but im already of the opinion that we should just remove it and rethink it entirely. thats why im not a fan of these "Middle Ground PRs" that invite scrutiny over a feature that should have been more scrutinized to begin with.

PR #2487, opened 2025-12-30
Gas bounties

GansuLalan:
## About the PR

Adding gas bounties, meant to do this ages ago but was trying to get something else to work and just petered out with doubts and stuff belggh.

This is currently only backend stuff, there was intention to add the related bounties but there's some issues regarding balance and transportation due to how bounties can only consider one storage option at a time. I'm just wanting to get this done so I don't have a draft floating around in the background for however long.

## Why / Balance

Last bounty type that needs adding unless we feel like getting cargo to be assembling furniture (I do love a good factory) and I want to make atmos make the funny gases. The addition of bounties themselves have been shelved until I can figure out a better solution for the transportation issue.

## Technical details

Added a new bounty type same way as reagents. This includes some ftl stuff that's a bit redundant but will probably be used at a later date so I'm leaving it behind, also commented some stuff in canisters since I was toying with labels but dropped it for the aforementioned reasons.
All changes can be considered backend and have no effect on current gameplay unless I missed something but it should all be fine from what I have seen.

## Requirements

## Breaking changes

**Changelog**
No changes to gameplay so no patchnotes but for the sake of clarity:
Added the code for gas bounties, they can be created via using the gas tag when defining a bounty as per the gas enum
Added the ability to set target limits for categories e.g. X category can only ever have 2 targets generate at most via the maxTargets field (default is "infinite")

### DIFF
--- a/Content.Server/Cargo/Components/StationCargoBountyDatabaseComponent.cs
+++ b/Content.Server/Cargo/Components/StationCargoBountyDatabaseComponent.cs
@@ -65,5 +65,5 @@ public sealed partial class StationCargoBountyDatabaseComponent : Component
     /// The time between skipping bounties.
     /// </summary>
     [DataField]
-    public TimeSpan SkipDelay = TimeSpan.FromMinutes(0);
+    public TimeSpan SkipDelay = TimeSpan.FromMinutes(3);
 }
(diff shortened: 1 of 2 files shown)


taydeo (approved):
how did this slip thru

PR #2545, opened 2026-01-03
Gondola Fix for being brainless

DimFlower:
## Why / Balance
Gondolas not dropping brains was a bug. This fixes the bug and slightly dampens how powerful of a Round-removal it was, and allows some MINOR continuance of play after. 

## Technical details
The old Gondola file overridden and renamed, and a new Gondola file added to funky, nigh identical. The only change *yet* is an inheritance change, though more intended to come in future. 

## Requirements

## Breaking changes
This will be posted in #codebase-changes. --> Gondola Body Inheritance changed to Primate

**Changelog**
:cl: fix: Fixed Gondola organ drops and RR


A-Loose-Goose:
I thought the whole point of Gondolas was to RR? You take enough Tranquility and you're turned from a guy to a thing. There's no point in buying Gondolas if they have no utility. Not a huge fan of this.

Komonighub (commented):
yeah this is a lot of yaml movement, taydeo id look at this since idk why/if we need to keep the old entity or if we want gondolas to drop brains at all since it removes the point of gondolaing

PR #2555, opened 2026-01-04
More Job Titles!

SharkSnake98:
## About the PR
Adds multiple new Job Titles, including those which previously lacked any, with the exception of trainee jobs.

## Why / Balance
More customization is nice, especially for Heads!

## Technical details
YML and FTL

## Media
image
image
image

## Requirements

**Changelog**

:cl:
- add: Added a load of new alternative job titles!


vectorassembly:
I think we specifically avoided having job titles for head roles.

Teasq:
I don't think heads should have alternative titles. It's not as important for less important roles, but for accessibility and quite frankly realism purposes heads should be immediately identifiable by their job titles. 

It's easy to say people are more comfortable with it now, but I'm not sure that's actually a verifiable statement. New players should be able to identify some of the most important people on the station easily, and not be subject to the 15 or more titles they now have to learn to differentiate between them.

PR #2567, opened 2026-01-04
Paper may now be stamped up to 10 times

TrixxedHeart:
## About the PR
Allows for paper to be stamped visually multiple times up to 10 times.

## Why / Balance
the funny
10 was chosen as the max because its still hilariously excessive without potentially causing issues. You can still try to stamp more than 10 times but only the first 10 will be visible.

## Technical details
Changes a few lines in `PaperSystem.cs` to allow for multiple of the same stamp.

Also fixes an oversight in `PaperWindow.xaml.cs` where stamps would always make a new line regardless if it's already made space for being stamped, which should no longer lead to issues like this:

https://github.com/user-attachments/assets/b390b50c-5edc-4b70-86af-6ef74ba3521c

## Media

https://github.com/user-attachments/assets/faacb89e-9a50-4fe6-9a6e-7ff12ff3f4d7

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: You can now stamp the same paper multiple times with the same stamp (up to 10 times)
- fix: Stamps now only make space for stamps the first time a paper has been stamped, instead of making a new line every time it's been stamped.

### DIFF
--- a/Content.Client/Paper/UI/PaperWindow.xaml.cs
+++ b/Content.Client/Paper/UI/PaperWindow.xaml.cs
@@ -19,6 +19,7 @@
 // SPDX-FileCopyrightText: 2025 jackel234 <52829582+jackel234@users.noreply.github.com>
 // SPDX-FileCopyrightText: 2025 pathetic meowmeow <uhhadd@gmail.com>
 // SPDX-FileCopyrightText: 2025 taydeo <td12233a@gmail.com>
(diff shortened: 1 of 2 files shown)


taydeo (approved):

Dragon232347:
I would change the changelog/title for this to "you can now use the same stamp on the same paper up to 10 times" because rn it sounds like you're limiting the total number of stamps to 10

PR #2606, opened 2026-01-08
Cosmic Cult Tweaks

ferynn:
## About the PR

Two small changes, but might change the flow of a cosmic cult round some.

The effective entropy of a cultist was changed from 7 to 3 for conversions.

The absorb time for malign rifts was changed from 35 seconds to 20 seconds, equal to the chaplain's speed of dispersal.

## Why / Balance

The goal here is to lengthen the time it takes for stage 2 to happen, and allow for more conversions, more entropy, and so on.

The cult is often a bit too weak currently when security mobilizes to hunt down the threat, so this will hopefully give them some more time to grow, and perhaps encourage other ways of advancing the counter, like the entropy generators.

The absorb time was changed to make it more feasible for cultists to actually get any of the rifts. Often they are swarmed and watched, so I think it's not unreasonable to make them at least as fast at the Chaplain in absorbing them.

These changes are small and exploratory.  There will likely need to be further changes.

## Technical details

Two number changes in DCCVars.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Cosmic cultist conversions are worth less entropy to the monument
- tweak: Cosmic cultists can absorb rifts faster.


Komonighub:
@taydeo balancing pr

PR #2655, opened 2026-01-12
Removal belt loadout option from NTR

TheHolyAegis:
## About the PR

I removed the belt loadout option.

## Why / Balance

The NTR already spawns with both the clipboard in the bag and the waistbag on the belt. There is no point to have this option here. And I noticed if you select either of them it will drop on the floor instead when you spawn in.

## Technical details

YAMLing

## Media

N/A

## Requirements

## Breaking changes

None

**Changelog**

🆑 
- remove: Removed the belt loadout option from the NTR, due to it already spawning with the waistbag and clipboard.


Teasq:
@taydeo NTR loadout fix

PR #2678, opened 2026-01-14
[Wizden Port] Gibtonite ore 

ThatOneMoon:
LICENSE: MIT
## About the PR
Ports over [ #37160](https://github.com/space-wizards/space-station-14/pull/37160) from Wizden

Adds gibtonite, an explosive threat, into mining asteroids and the vgroid.

This rock will prime once you hit it, giving you only a few seconds to run away before it explodes.
The blast is decently big and will shred nearby rock, but is also capable of nearly critting a salvager who gets stuck in its blast.

It's not visible until you hit at, at which point you have to quickly flee.

## Why / Balance
Asteroid mining itself is rather empty of threats or interest. I don't really want to just add tons of mobs into it, as that makes it a bit too much like regular scrapping and whatnot, so i find this to be a good compromise.

Gibtonite serves a decently weak threat

## Requirements

**Changelog**
:cl: ThatOneMoon, EmoGarbage404
- add:  Added gibtonite! This secret ore becomes visible when struck and explodes violently. It can be detected early using a mineral scanner. (Originally by EmoGarbage404 from Wizden)

### DIFF
--- a/Resources/Changelog/Funkylog.yml
+++ b/Resources/Changelog/Funkylog.yml
@@ -7602,3 +7602,13 @@ Entries:
       message: Removed the old atmos spawners and replaced them with engineer spawners
   id: 879
   time: '2026-01-30T10:19:50.0000000+00:00'
+- author: Huhm-1
+  changes:
+    - type: Add
+      message: Added Packed to the map pool
+    - type: Remove
+      message: Removed Amber from the map pool
+    - type: Remove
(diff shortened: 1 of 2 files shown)


taydeo (approved):

Harro365:
This both makes mining more risky, more reliant on sci, and more fun. I hated this feature on Wizden as much as I loved it, and this seems like a perfect port.

PR #2686, opened 2026-01-16
Suppressed Bodily Cues Quirk

Chedd-Error:
## About the PR

Added in the suppressed bodily cues trait that makes it so that you do not get the peckish or thirsty alerts when your hunger/thirst is low respectively. Also made it so that more alerts can be defined for the thirst component.
## Why / Balance

We have a plethora of quirks that seem to only negatively affect your characters, and I think that this would be a fine addition to said roster. While it may lead to characters not seeking out bar/kitchen RP as they only do enough to keep their statuses from reaching a critical level, I feel that it could lead to the contrary; your character notices that they have not had anything all shift and all of a sudden feels a sudden wave of hunger sweep over them, and might seek a proper meal instead of trying to fill up on snacks. As for the thirst component changes, that was a change I made during an earlier approach at implementing the trait which brings the thirst component's fields more in line with the hunger component's, and may be of use if we ever decide to implement an antagonist/species with special focus on being well-hydrated.
## Technical details

Exposed the ThirstThresholdAlertTypes to be editable via the Thirst component under the field thirstThresholdAlerts
Added in if-else clauses to the part of the code that handles alerts in the ThirstSystem.cs and HungerSystem.cs files.
Added in a file under the Content.Shared/Traits/Assorted to handle the IgnoreMinorBodilyAlerts component
## Media

https://github.com/user-attachments/assets/3e63dc7e-8c94-4b62-a3a5-12a3411b5cfd

## Requirements

## Breaking changes

**Changelog**

:cl: zergologist
- add: Suppressed bodily cues quirk, don't forget to eat and drink before your body kneecaps you for it!


Teasq:
@taydeo new trait, suppresses food cues until you're starving.

PR #2687, opened 2026-01-16
Synth trait actually can't detect ion storms now

mycobiota:
## About the PR

RobustRandom.Prob, the function used here to decide whether a synth character should get an ion storm alert, returns True if a random number between 0 and 1 is less than the passed "chance" value
Originally, the logic flow meant that SynthComponent's AlertChance doesn't specify how likely you are to get an alert, it specifies how likely you are to not get an alert
In other words, the outcome of the random check has been negated, so where before AlertChance = 0 means 100% chance of ion storm alert, it now means 0% chance of alert
## Why / Balance

https://github.com/funky-station/funky-station/pull/2485
Synths detecting ion storms was meant to be disabled but because of this confusing logic they were given a 100% chance instead... so I fixed the logic
## Technical details

one line csharp
## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Characters with the synth trait actually can't detect ion storms now.


otokonoko-dev:
Requesting review from Taydeo.

Minor bugfix

PR #2690, opened 2026-01-16
Remove Chloral Hydrate injector from Shiva, Arachnids, and Spiders

ferynn:
## About the PR

I removed the arachnid/spider sleep bite ability.

## Why / Balance

The ability is overpowered, regardless if it costs hunger or not. It's a hypopen of sleep chemicals, even if delayed. One activation will cause the target to fall asleep multiple times, something that is crippling to antags, especially solo ones. The fact that is causes hunger is irrelevant, as you only need to use it once.

I don't think Shiva or spiders should have this. The idea of injecting types of toxins is a fun one, but they should be more traditional ones. Not chloral hydrate.

Shiva especially doesn't need it, as she does sec work. If it seems a bit odd conceptually to give a secoff a hypopen of chloral hydrate passively, it's probably not good for Shiva to have it either.

## Technical details

Deleted action from yaml

## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: Removed the chloral hydrate bite from arachnids, Shiva, and Tarantulas.


Drywink:
Saying it there too, cause i think its a neat idea :
What about, removing the biting ability, but instead having a small chemical storage in the fangs, that you can fill with whatever you want, and whenever you use the bite attack, it inject a small amount of that reserve into the bloodstream of the target?

taydeo (approved):

PR #2705, opened 2026-01-18
No more pipestacking

TheHolyAegis:
## About the PR

This PR removes the ability to connect a straight pipe to a corner pipe on the same tile (as an example. It also prevents connecting to other devices.).

Now, when you try placing a pipe with the RPD, it immediately unanchors when placed, the message "The {device} doesn't fit over the other pipes!"still appears. If you try anchoring it with a tool it immediately cancels the attempt. And if you construct it manually on that spot, it immediately unanchors upon completion.

## Why / Balance

Well, I have multiple reasons to change this:

- Firstly, there was always the message "The {device} doesn't fit over the other pipes!" Though, you could still place them like that.
- Secondly, I personally always found it disturbing to see, it doesn't feel realistic to see pipes setup like that (not that SS14 is realistic...)
- Thirdly, it can get pretty confusing to see when you are not the one that placed it.
- Fourthly, some sprites go z-fighting, you can see that now especially now the sprites are changed, it wasn't so bad when it was white on white, which is not really nice for the people with epilepsy. 
- And finally, atmosians already get three spots on one tile to place their pipes, they do not need to stack them weirdly. In some instances they could literally use a T-pipe or a four way pipe. It isn't hard to deconstruct a pipe to then change it.

## Technical details

I changed a boolean to true, the code was already well written, just turned off

## Media

image

No more stacking like this, I tried adding the corner pipe last with an RPD, and it's unanchored as you can see.

image

You can still use every spot on one tile like this

## Requirements

## Breaking changes

Maybe if some maps contain setups involving pipestacking will be weird.

**Changelog**

🆑
- remove: Removed the ability to stack the ends of pipes.

### DIFF
--- a/Content.Shared/CCVar/CCVars.Atmos.cs
+++ b/Content.Shared/CCVar/CCVars.Atmos.cs
@@ -3,6 +3,7 @@
 // SPDX-FileCopyrightText: 2025 VMSolidus <evilexecutive@gmail.com>
 // SPDX-FileCopyrightText: 2025 corresp0nd <46357632+corresp0nd@users.noreply.github.com>
 // SPDX-FileCopyrightText: 2025 taydeo <td12233a@gmail.com>
+// SPDX-FileCopyrightText: 2026 TheHolyAegis <sanderkamphuis719@gmail.com>
 //
 // SPDX-License-Identifier: MIT
 


rottenheadphones:
As much as i agree with the unrealistic part, i believe this deserves a discussion in either #development or #atmos/engineering general thread

PR #2712, opened 2026-01-20
FELINID NERF

BrightNibbleston:
## About the PR

increased how fast felinids get thirsty by 0.02, decreased their un-armed melee damage by 0.5 piercing, and reduced their heat capacity by 10, and increased how much cold damage they take by 0.25.

## Why / Balance

felinids are too strong as they are currently, like. they should be lighter and therefor less dense than humans so i reduced their heat capacity so they get cold faster to represent their reduced mass. their claws also wouldnt slash so deep as they would far so i decreased their un-armed piercing damage by 0.5. and cats usually are covered in fur but felinids are not so they should be weaker to cold by a decent bit. so i increased the cold damage they take to 0.75. also increased how much water they need due to all the meowing they do

## Technical details

changed a few numbers in felinid.yml

## Requirements

**Changelog**

:cl:
- tweak: nerfed felinids!

### DIFF
--- a/.github/workflows/labeler-untriaged.yml
+++ b/.github/workflows/labeler-untriaged.yml
@@ -4,6 +4,7 @@
 # SPDX-FileCopyrightText: 2024 Vasilis <vasilis@pikachu.systems>
 # SPDX-FileCopyrightText: 2024 slarticodefast <161409025+slarticodefast@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2025 taydeo <td12233a@gmail.com>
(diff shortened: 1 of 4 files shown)


TheHolyAegis:
Felinids were never too strong, I have never seen someone complain once about them, and they already get thirsty fricking fast, you want people to not play them? And you forget that cats have fricking fur that prevents them from getting cold quickly, how does that work?

TheHolyAegis:
Yeah, I already had noticed that. I used to two shot a mouse, now it takes like 5 slashes

PR #2714, opened 2026-01-20
commented out CCI spawnpoint on centcom.yml

DeltaSS14:
## About the PR

Removed CCI as a roundstart ghostrole
## Why / Balance

There's been plenty of discussion around CCI as a ghostrole but most of it is pretty scattered. Others and myself didn't realize CCI was becoming round start due to overlooking the PR that added it, hence the mald PR here instead of a comment before it was merged.

Messages from Central Command are a big deal. They hold nearly ultimate power, being able to overrule the Magistrate and NTR (and with that, pretty much any role on the station.) From a narrative perspective command should be terrified of Central getting involved, they are likely overseeing multiple stations and their eyes on you most certainly means you're doing something wrong. What helps with this is that they are rare, getting a fax back is very uncommon, which further props up their importance.

In game, we've been able to treat communication from Central Command like a big deal because up until now it's always been an admin spawn. Meaning that the message is either written by an admin themselves, or by someone who's been allowed to play a CCI by ahelping, who is thus endorsed by an admin to play one and/or has direct admin oversight. This gives players the OOC assurance that any communication from Central Command is accurate, valid and appropriate to the situation.

CCI as a round start ghostrole removes this assurance, and in it's current implementation leaves a lot unclear. I would propose keeping the role as an admin only spawn.

If there **is** a desire for a round start CCI ghostrole I would like to still temporarily remove it until it's more clearly implemented, for which I propose one of the following two solutions:

1. Write clear rules about that a CCI can and cannot approve and how they should behave in the guidebook and Rule 15. Keep their power above the Magi and NTR but seriously raise the playtime requirements. (I would suggest either 50 hours of magistrate as that requirement itself involves a lot of time in high-responsibility roles, or 100 hours in central command)
2. Write clear rules about that a CCI can and cannot approve and how they should behave in the guidebook and Rule 15. Do not allow them to overrule or order around Magi and NTR and keep the current playtime requirements. Make it possible for the station to differentiate between a CCI and a CCO by giving them different stamps. [EDIT] Additionally remove the holocall and have their communications console access revoked so they communicate only by faxing.
## Technical details

Commented out the CentComIntern spawnpoint on _Harmony/centcom.yml [EDIT] removed instead of commented out
## Media

None.
## Requirements

## Breaking changes

**Changelog**

:cl: Delta
- remove: Central Command Intern as a ghostrole at the start of the round

### DIFF
--- a/Resources/Maps/_Harmony/centcomm.yml
+++ b/Resources/Maps/_Harmony/centcomm.yml
@@ -41,6 +41,8 @@
 # SPDX-FileCopyrightText: 2025 slarticodefast <161409025+slarticodefast@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2025 taydeo <td12233a@gmail.com>
 # SPDX-FileCopyrightText: 2025 wilowzeta <willowzeta632146@proton.me>
+# SPDX-FileCopyrightText: 2026 unknown <kit.ds.kat@gmail.com>
+# SPDX-FileCopyrightText: 2026 willow <willowzeta632146@proton.me>
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later
 


PoquitoDondito:
I’ve written my thoughts up in a slightly more inflammatory way (to be clear, I was mostly playing it up) than I would if I was just typing them here, but I don’t have time to rewrite everything so I’ll just post it:
IMG_0350
I’ll say in addition to this is that I personally want to see a couple things first before we implement CCI:
1. OOC rules/guidelines/sop for CCI, that way players clearly know what is expected of them.
2. A CCI stamp, which they are required to use every time they send a fax.
3. Magistrate and NTR time, since they need that experience to be able to overrule the NTR and magistrate on station.
4. OTHER ghost roles at CC, so that CC staff doesn’t just order executions because they’re bored (relatively minor point).
5. Explicit statement in rules that CC is subject to EORG rules throughout the entire round. 

I think that’s pretty much it, There might be more things that come to me but I wanted to make sure I at least contributed with that first.

KittyCat432:
Can we just shoot them to be normal :3 

But seriously if there was some way this could be implemented as a actual ghost bar replacement that would be amazing. You die permanently 
Oh theres some people on CC let me join as thier medical person and talk in the bar

This would be amazing if only people could just be nice so we can have nice things.

Idem778:
> I actually do like the idea of CCI and want to move in the direction of having them be a round start ghost role

Quick thing, this already is the case. CCI is a roundstart ghostrole and it completely randomizes the character with a random name and appearance and even species

DeltaDnDsolutions:
> The ghost roles CCI I've seen from this update have grossly misused/overused the communications console, made poor remarks, power tripping, and overall have been a net negative on roleplay experience. The fact alone this was done without creating any SOP for them or the likes makes it much worse.

Adding onto the option 2 that I described in my post, it would be my least preferred option, but if that is the way things are taken then I'd like to see their access to the communications console and the holocall removed, making it fax-only communication.

> The only way I could _ever_ realistically see CCI being a role outside of admin spawns is if it were whitelisted. It's a role that has very specific expectations from station players and not everyone is fit for it just because they have 20h CC.

I can agree with this.

Idem778:
> > > I actually do like the idea of CCI and want to move in the direction of having them be a round start ghost role
> > 
> > 
> > Quick thing, this already is the case. CCI is a roundstart ghostrole and it completely randomizes the character with a random name and appearance and even species
> 
> Yeah I know it is, that's just about the only thing I like about the current implementation :P. There's just been some discussion that we should make it a selectable job and I disagree with that for the listed reasons.

Ah sorry! Didn’t read that!

TheHolyAegis:
It was fun while it lasted, I think I was a great CCI. I am definitely going to play it some more before the server gets updated

DeltaDnDsolutions:
> There can be good CCIs, but also people who abuse it. People have common sense about how to play CCI, even without rules. But yeah, you can also just ahelp to ask to become CCI when they get removed again.

This is what rules and guidelines are for, so people who don't have the common sense to play the role as intended have something to follow, and can be corrected if they do it wrong. But even with that, given their immense power and how hard it can be for admins to oversee them, see: 

> Only problem is that it's an entirely separate map, so it's much harder for admins to monitor both maps at once, since you have to teleport between them and can only hear what's happening on one map at a time (unless it's CC comms)

I'm starting to agree with Trixxed more and more that it should either be a whitelisted role only (which has been mentioned is likely never happening) or admin approved only so they can approve it when they actually have the time/energy to babysit a CCI if needed.

PR #2716, opened 2026-01-20
Ghost job preferences + CCI All Access

TheHolyAegis:
## About the PR

The less important one. I gave CCI all access.

And, the more important change. I made it possible to use your own character for certain ghost roles. Currently only CCI and Space Cowboy are supported (I think, maybe the visitor ghost roles work with it too.).
I have added a new "department" section called "Ghost Jobs." Currently only CCI and Space Cowboy are listed there, but people can add more. When you select the ghost role when you are an observer, it will pick one of the characters you have set it to yes with. If no characters have it set to yes, it will create a random character like it did in the past.

## Why / Balance

For CCI getting all access, it's because during the moment no faxes come in, you might want to stroll around and just do stuff. Though CCI can't enter the departments because they couldn't access it. And I gave it all access because I think it would make sense, could also work nicely if they get sent to the station by an admin, less setup time.

And for the other change, I thought it would be nice if you could use your own character as CCI, space cowboy just get added because it worked out. And now since there are always 2 intern ghost roles roundstart it was time to add it.

## Technical details

I remade the spawning script of the ghostrole jobspawners, since those only spawn a character when someone selects the ghostrole. I added some checks and it works. If you want to look in depth, you can look between the changed files at the GhostRoleSystem.cs file.

I added a field called VisibleInLobby which is true by default, but when false it removes the icon from the job priorities screen.

## Media

image

All the way at the bottom you have ghost jobs now.

image

They are not between the icons here.

## Requirements

## Breaking changes

There are no breaking changes

**Changelog**

🆑
- tweak: Gave Central Command Intern all access.
- add: Made it possible to use your own character for certain ghostroles. You can enable them in the jobs screen of your character.

### DIFF
--- a/Content.Server/Ghost/Roles/GhostRoleSystem.cs
+++ b/Content.Server/Ghost/Roles/GhostRoleSystem.cs
@@ -107,7 +107,6 @@ public sealed class GhostRoleSystem : EntitySystem
     [Dependency] private readonly IServerPreferencesManager _prefsManager = default!;
     [Dependency] private readonly StationSpawningSystem _stationSpawning = default!;
     [Dependency] private readonly StationSystem _stationSystem = default!;
-    [Dependency] private readonly IServerPreferencesManager _preferences = default!;
 
     private uint _nextRoleIdentifier;
     private bool _needsUpdateGhostRoleCount = true;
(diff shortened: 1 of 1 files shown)


TheHolyAegis:
This could still work even after the roundstart CCI ghost roles get removed, for if the admin decides to spawn some in.

This only works for the ghost role spawners, not when the actual character is placed down like ERT are added

TheHolyAegis:
Could still be nice for other ghost roles if they get added. And I thought I heard that CCI is getting removed to then make some better rules and stuff for them

PR #2721, opened 2026-01-21
Constructable Graves

Xcybitt:
LICENSE: MIT
## About the PR
Made steel graves(literally just renamed stone graves but a new prototype) and wood graves craftable with steel and wood respectively. Since graves are not deconstructible and they need 200 damage to actually break, made it climbable. Graves also take a long time to dig, so hopefully it isn't gonna be used as a quick cover option. Also labeled the shovels and the like that can dig as having the digging tool property, put the shovel recipe into lathes, and gave the chaplain a shovel in their vend. And yes, even if you have no shovel, you can still claw out of the grave(Code was like that already, which is neat).

## Why / Balance
Graves are cool, and corpse management RP is fun! Burial would be more of an option with this, and maybe you can bury corpses in mass graves as a temporary mass burial thing. Antagonists could also bury their victims alive or dead. You also could theoretically have a dead coming back to life thing for lings, zombies, and the like that revive.

## Technical details
YAML 

## Requirements

## Breaking changes

**Changelog**
:cl: Xcybitt
- add: Made graves craftable with steel or wood. You can jump over graves, too. Remember to bury corpses in graves so they can't get back up! Maybe. Label the grave with the victim's names on a paper so you don't forget!
- add: Added the shovel to lathes and the chaplain vending machine!


Xcybitt:
> was this at all inspired by the incident image

the incident literally is the one thing that reminded me that graves exist, so yes

Xcybitt:
i didn't touch trait code 😭

PR #2727, opened 2026-01-22
Disablers now only do slowdown

TrixxedHeart:
LICENSE: MIT
## About the PR
Removes stamina damage from anything that uses the `BulletDisabler` projectile and replaces it with a slowdown.

## Why / Balance
To preface, I'm a Security player.

Long-range, low-commitment stuns are boring. Disablers currently function as a one-size-fits-all solution for Security, allowing officers to fully incapacitate targets at range with minimal risk, positioning, or coordination. This reduces the need to engage with Security’s wider toolkit and makes encounters predictable and frustrating for the target, with also little to no methods to prevent it.

By removing stamina damage from BulletDisabler projectiles and replacing it with a slowdown, disablers retain their role as a chase and control tool without being a complete solution on their own.

**Current scenario:**

- A Security Officer attempts to arrest someone in a hallway.
- The suspect runs.
- By the time the officer reacts, the suspect is out of baton range.
- The officer defaults to the disabler, landing a few shots and securing a ranged stun with little counterplay.

**This change:**

- The disabler remains useful for applying pressure and preventing a clean escape.
- Actual stunning now requires closing distance, baton use, or coordination with another officer.
- Officers must commit to positioning and timing rather than relying solely on qolslop stun guns and actually beat the shit out of people with their batons.

Ultimately this
- Encourages security to utilize other tools in their toolset while keeping the disabler a useful tool.
- Enfoced distinct uses, disablers slow and control, batons finish.
- Reduces Security’s ability to safely end encounters from a distance (no more just fucking stunning the nukie to shut them down instantly)

Overall, this preserves the disabler’s usefulness while preventing it from overshadowing melee stuns and actual skill. It shifts the focus from ranged shutdowns and toward more deliberate play.

## Technical details
Comments out `StaminaDamageOnCollide` in favor of `StunOnCollide` which will ONLY slowdown the target instead of any stamina damage. This is capped to only a max of 50% to prevent it from slowing people to a complete stand-still.

Adds `SlowdownCap` and relevant code in the `StunOnCollideSystem` to enable the ability to cap slowdowns.

This does not remove stunning from turrets.

## Media
I cant be assed to try and fit the videos into github's file size so just have this
https://discord.com/channels/1276640157511979008/1276653734767755294/1463971041188515942

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Disabler Bolts no longer do stamina damage, and now only apply slowdown.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -75,6 +75,7 @@
 # SPDX-FileCopyrightText: 2025 pa.pecherskij <pa.pecherskij@interfax.ru>
 # SPDX-FileCopyrightText: 2025 slarticodefast <161409025+slarticodefast@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2025 taydeo <td12233a@gmail.com>
+# SPDX-FileCopyrightText: 2026 Mora <46364955+TrixxedHeart@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2026 ThatOneMoon <91613003+ThatOneMoon@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2026 ThatOneMoon <juozas.dringelis@gmail.com>
 #


Harro365:
I preface by saying this: I am a sec main
I have something like 1.8k hours in the department across servers, sec is in my blood. I understand what this PR aims to fix, the age-old stun meta debate; however, it is my opinion that this, in fact, does not fix jack shit. I love sec, but also I also love positive change and good game design more than habit.

I am simultaneously addressing that this PR aims to fix a real issue, and also arguing that it aims to fix this issue in the TOTAL WRONG WAY. Here's why:

Disablers are, as of now, perfect for their role. How I see it described here appropriates the tool and phrases it to sound as if it does not have 90 counters, and is also very limited in usecase. It honestly creates an issue. Disablers are not a replacement to closing distance, nor the best tool for arrests, nor a replacement for a baton. The disablers' use is providing a method of incapacitation at distince AND slow without crossfire. It also requires several consecutive hits (ignoring fellinds) with little to no time delay, open space, and line of sight. This PR phrases disablers like how they are on RMC rather than the reality. 

You still have to close the distance as a secoff; the disabler is not a QOL-slop substitution for skill. It is also necessary for many kinds of arrests, when you do not want to escalate to lethal force but also can not get into baton range. Having it just slow kills the usefulness so much as to make it so that it is not worth taking. If you need slow, you already have the flash and bola to do it MUCH BETTER. 

The disabler stun meta is already easily countered. As a sindi, using a eweapon of any kind makes the number of shots required for a full stun exponential. It is also easily countered by an EMP. As any antag, just using basic movement strategy, understanding line of sight, and how to counter it is all that is needed. Antags have SO MANY TOOLS to deal with a braindead disabler heavy secoff. It just takes an understanding of how combat works and how to use space and light.

For melee, a baton is ALWAYS BETTER. It is faster, easier to use, and harder to counter, provided the officer understands slip meta and disarm mechanics. Making the disabler stun does not help the baton; it needed no help. It was already perfect in its role, as was the disabler. All this would do is make officers have to reach for their lethal option instead of dealing with threats they could otherwise have followed ROE with a disabler, and made countering easier for the antag to promote mutual skill.

Making arrests nightmares is not QOL slop. It is compensating for antags opting not use strategy and skill. Removing the disabler stun then is just antag QOL slop and adding the disabler to the toolset that the flash and bola already does better. 

Running someone down with a baton after slowing them is in no way a skill increase. To say that it is is in reality just wrong. I would argue that it takes even LESS skill to baton someone. Slowing to 50% also means you just chase until they run into a dead end then disabler, then run down to baton into stun and cuff. 

Again, it is not like the baton NEEDED THE HELP. It is not ignored. It is used as much as it should be.

THIS IS IGNORING THREE OTHER THINGS THAT NEARLY ALL ANTAGS GET
Chems, stun res armor, and EMPs.

Changelings get infinite ephi. Any sindis can buy hyperzine, the end-all be-all counter to disablers (botany can make it too with a bit of effort). Chem or botany can make ephi or desox easily. Sindis, changelings, and chem can all make EMPs to disable every disabler (I am ignoring the 1/10000000000 chance of sci getting a handheld with EMP and an easy activation like auto. Antag armor from Bloodred Hardsuits to changeling armor to Heretic armor all get some kind of stun resist that stacks to add 2-4 extra shots in a row needed. In fact, nearly every combat armor set does! 

Now, my goal is not to say why I hate this thing, but to say why I love it, and I do! I agree that disablers can feel too easy! Here are ways I would say to ACTUALLY fix the issue without just making more issues.

- Slightly lower the damage to require another 1-2 hits-in-a-row. This promotes good aim and follow-through, and rewards antags for good mobility and space utilization.
- Slightly lower the fire rate so shots have to count more, promoting aimed hits rather than skillless spray-and-pray

Both of those would, in my opinion, be more effective at this goal.

PR #2771, opened 2026-01-30
Temperature-based fire color variations

vivry:
## About the PR

This is a minor addition to the recently-added reagent fire system. Now, there are three variants of all reagent fire sprites that change depending on how hot the fire tile is. Fires burning at <1000C are now more red, while fires burning at >2000C are more yellow than the default texture. This doesn't affect the big fires (ones that take up an entire tile)

## Why / Balance

I think that this is a small flavor addition that is both realistic and useful in gameplay, since it gives the players an ability to visually gauge how dangerous a fire is.

## Technical details

The implementation is very simplistic - instead of making major changes to gas overlay drawing, I added more animation frames to the fire textures. Depending on the fire temperature, the client plays a subset of the frames that represent the sprite burning at the correct temperature. A more thorough and robust implementation would involve finding a way to recolor the sprites at runtime (I tried a few ways but none of them looked good), and adding more color variations for a more gradient-like look.

## Media

Fire Colors

## Requirements

## Breaking changes

**Changelog**

:cl: vivry
- add: Added reagent fire color variations that depend on the fire's temperature


YaraaraY (approved):
looks good to me! i like it!

taydeo (approved):

taydeo (approved):
i remember there being a discord discussion so if @Komonighub can ask the authors about it That'd Be Great

otherwise im ok w it being merged as is

YES, I KNOW, EVEN WITH THE HARD CODED GAS TEMP STUFF

PR #2776, opened 2026-01-31
Allow documents to be created from text files

beck-thompson:
licence: MIT (standard MIT)

from (https://github.com/ss14Starlight/space-station-14/pull/2295)

## Short description

Conf wanted it. FTL files are very painful to write complex documents in so this seems like a reasonable solution

https://github.com/user-attachments/assets/73606b8b-baf2-484c-bee8-32701b7e60db

some example prototypes to test it:

```
- type: entity
  id: TextLetterOne
  name: text letter one
  description: description 1
  parent: Paper
  components:
    - type: Paper
    - type: TextFilePaperContent
      fileName: test_document_1.txt

- type: entity
  id: TextLetterTwo
  name: text letter two
  description: description 2
  parent: Paper
  components:
    - type: Paper
    - type: TextFilePaperContent
      fileName: test_document_2.txt
```

:cl:
- add: Documents can now be created with text files


ParadoxumOfficial (changes requested):
Please update the description to include a working change log, it is required.

taydeo (approved):

PR #2844, opened 2026-02-07
washing machines fucking hurt you now

YaraaraY:
## About the PR
This PR removes the whitelist from the washing machine contents, and deals damage over time to entities inside when the machine is turned on. It deals about 30 blunt damage in total, and then triggers a migraine in the entity to simulate dizziness. The washing machine also deals damage to itself, which will be enough to break it after 2 or 3 cycles if you keep shoving people inside. It also sprays the contents inside with 150u of water, meaning if you shove a slime inside they're gonna melt.

## Why / Balance
This is very fucking funny.

## Technical details
Some C#. Shouldn't be anything too egregious.

## Media
https://github.com/user-attachments/assets/7f5c783a-c349-4478-a5ed-e49c8778cc50

## Requirements

**Changelog**
:cl: AraiMaia
- add: You can now shove people inside washing machines, who will deal a total of 30 blunt damage and splash water to whoever is inside, while also making them horribly sick.
- tweak: Tweaked the washing machine behaviour to be more intuitive to how other containers work in-game.


Xcybitt:
mm maybe a bit of feature creep but theoretically could you make it so that this also works for the autoclave and make the  washing machines and the clave vaultable?

YaraaraY:
> mm maybe a bit of feature creep but theoretically could you make it so that this also works for the autoclave and make the washing machines and the clave vaultable?

would be a different pr

PR #2828, opened 2026-02-06
Security Redux

pirakaplant:
LICENSE: MIT

(Or: _How I Learned to Stop Worrying and Harmbaton Prisoners_)

## About the PR

This PR is part of an effort to enact the changes proposed by funky-station/docs#60. Disablers and stun batons have been removed entirely from normal gameplay. (They are still in the files itself as prototypes, but they have been removed and/or replaced with alternatives across loadouts, locker/crate/safe contents, and lathe recipes. Mapped items have also been adjusted through prototype migrations.) Stun batons and disablers can no longer be lathed. Similarly, multi-setting laser weapons (the X-01 Multiphase, RG-07, and RG-4) no longer have a disabler setting.

Nightsticks (formerly known as truncheons) are now available roundstart, albeit in a slightly weaker form (12 blunt per swing, from 20, and 5 when retracted), and knock down a victim on four hits. They fill a 4x1 area on inventory grids when extended, shrinking to 2x1 when retracted. In addition to their new name and statistics, nightsticks also have new sprites by @YaraaraY and sounds edited by me and @IamJustHereYknow.

Jobs that would start with a stun baton (such as security jobs) instead start with a nightstick. Lathing nightsticks no longer requires Advanced Riot Control to be researched. Some jobs that would start with a disabler and no lethal sidearm (such as the Nanotrasen Representative) start with an RG-2 instead. SOP has also been updated accordingly so that jobs that were allowed to carry disablers are now allowed to carry lethal sidearms.

There are also several other minor changes across the board to fully incorporate this change. For example, the practice disabler is now called the practice blaster, several references to "nonlethals" are replaced with "less-lethals", the heretic's Mask of Madness ritual uses a stun prod instead of a stun baton, and the Security entry in the guidebook has been updated accordingly.

(Note: originally, this PR used truncheons as they appeared prior to this PR. However, they have been changed to nightsticks following internal discussion.)

## Why / Balance

I wrote about this a bit in the design document (funky-station/docs#60), but I'm going to provide a full explanation for our vision of Security, for those out of the loop.

> people over the course of 20 years have essentially created a great analogy for class conflict and no server, from any i could see, wants the game to properly emulate it [...]
> — @taydeo

Lately Funky's been going in a more cohesive direction. Lately there has been a lot of discussion among designers and staff alike concerning the vision for what the server could be, and part of that is analysis of the inherent political themes that come with working for a questionable megacorporation in the far future.

[...]

PR #2855, opened 2026-02-08
Make NT Default the base lawset for Silicons

Rainbeon:
## About the PR
Required for #2828

## Why / Balance
Crewsimov Borgs and Security not having any non-lethals physically cannot co-exist, NT-Default is a good band-aid lawset at minimum (and also makes more thematic sense for our story direction).

## Media
Lawset for reference
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The Silicon default lawset is now NT Default, instead of Crewsimov.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Player/silicon.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/silicon.yml
@@ -56,6 +56,8 @@
 # SPDX-FileCopyrightText: 2025 puuupy <59433122+lolmatdi@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2025 slarticodefast <161409025+slarticodefast@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2025 taydeo <td12233a@gmail.com>
+# SPDX-FileCopyrightText: 2026 MaiaArai <158123176+YaraaraY@users.noreply.github.com>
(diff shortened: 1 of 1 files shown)


ilyamikcoder:
if i am reading this correctly, this does not apply to AI, only to borg chassis. i think it really should given the argumentation provided

SigmaTheDragon:
Do not forget that this would require changing how emagging borgs works too, else emagging borgs would likely lead to stuff like borgs refusing to steal an item because they have to protect the station and it's assets before obeying crew

YaraaraY (approved):
i think we should make a special, base lawset to allow for crewsimov shenanigans without every borg suddenly becoming enemies of sec, but for now this lgtm :)

taydeo (approved):

JohnnyDangerSpace4:
> This is in-fact temporary until a better lawset is made.

Why don't you just make a better lawset with the community and put the Security Redux on hold until then. It's not like you need to merge everything in a picosecond. It's been sitting there for two weeks and I'm sure you're not going to bleed security players like an open wound. Borgs laws have always been a really controversial thing and I feel like just doing a half measure here would just piss people off. Not that you really SHOULD care it is your server but its my own two cents.

Mot2332:
Small suggestion on the crewsimov solution, but how about :

 Law 1 :  you may not injure crew

 This would be in line with asimov's spirit, as it is the point in one of their stories how this version lets robots basically harm humans by not acting in dangerous situations.

PR #2868, opened 2026-02-11
Medical Scanning Is Kill

Xcybitt:
LICENSE: MIT
## About the PR
Commented out medtek's existence and also the handheld medical scanner's from most of the sources. Now, the CMO has one in their locker, and you can buy more in the emergency advanced medical kits. However, some special PDAs(Like the agent PDA) still has the medtek. Also added the bodyscanner board recipe to the circuit printer. Syndicate people also have combat kits where they can buy a medical analyzer.

## Why / Balance
Most damages are already viewable and guessable through the health examine, and honestly. Kinda all you need for high stress situations if everyone's shot, burnt or whatever. With this, medical will need to check damages on a patient in a scanning room if they're poisoned or whatnot, which I hope will make medical less of a just walk in and get healed sort of place, and give people more time to RP("Doc, what is it?" "You have cancer."). So, you can check it with a bodyscanner in the OR or in the cryopods.

## Technical details
YAML

## Requirements

## Breaking changes

**Changelog**
:cl: Xcybitt
- add: Added the bodyscanner computer board to circuit printers!
- tweak: Added some stuff to filled medkits to balance out losing the medical scanners.
- remove: Removed the medtek's existence alongside most of the sources of the handheld medical scanner. Now, you have to guesstimate your patients OR check their health in a cryopod or bodyscanner.

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Items/firstaidkits.yml
+++ b/Resources/Prototypes/Catalog/Fills/Items/firstaidkits.yml
@@ -23,8 +23,10 @@
 # SPDX-FileCopyrightText: 2024 slarticodefast <161409025+slarticodefast@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2025 DevilishMilk <michaellapjr@gmail.com>
 # SPDX-FileCopyrightText: 2025 Fenn <162015305+TooSillyFennec@users.noreply.github.com>
+# SPDX-FileCopyrightText: 2025 Mish <bluscout78@yahoo.com>
 # SPDX-FileCopyrightText: 2025 corresp0nd <46357632+corresp0nd@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2025 taydeo <td12233a@gmail.com>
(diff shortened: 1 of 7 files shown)


88tv:
> In short, this change seems like it'd be more fitting for HRP than MRP.

Important to note that Funky dropped the MRP tag on its server names weeks (if not months) ago. Try not to justify things on the basis of vague levels of roleplay.

Jack-Silver:
13's way of doing it I think is peak, ill boot up my private server after work to show off because I do think med scanners could be removed after some changes. So, on 13, the inspection details on yourself are way more detailed, e.g "My right arm is frozen to the bone. My left leg is bleeding profusely. My chest is slightly bruised. I feel sick to my stomach."

But then in medical, each bed is linked to a in depth medical scanner computer, that again, goes into way more detail, down to what's in the person's bloodstream, stomach, and the exact numbers of damage, if bones are broken etc. 

I think if we had those two systems med scanners would be pretty much redundant, and could easily be removed

PR #2872, opened 2026-02-11
Less Lethal Flashbangs

Xcybitt:
LICENSE: MIT
## About the PR
Since sec's becoming less lethal instead of non-lethal, why not make the premier nonlethal option also a less lethal. Flashbangs explode at around 1/10ths the strength of the normal grenade.

## Why / Balance
Parity to real life. Also just the fact that it feels weird that flashbangs can knock you over but it doesn't deal ANY damage. 1/10ths feels a little weak but it can be tweaked, worrying about people just stacking them tbh.

## Technical details
YAML

## Requirements

## Breaking changes

**Changelog**
:cl: Xcybitt
- add: Added the bang to flashbangs! They hurt now.


Xcybitt:
The explosion is like 3 tiles-ish in radius from the last I remember+at max it deals like 10 damage or sthn, it's really not much. However it can and will cook off other grenades.

taydeo (approved):

PR #2875, opened 2026-02-12
Command and CC are no longer rolled first 

willowzeta:
## About the PR

Removed the "weight" field in the prototypes of all Command and CC jobs. This makes it so that Command and CC are no longer prioritized in the selection process. You can now put those jobs on Medium or Low and actually roll something else!

Also, the commented out geneticist job had an improper weight, so I took the liberty of removing that.

## Why / Balance

With our job selection system as it currently is, you can currently put Command jobs on "low" and still get them practically everytime, since command jobs were selected at the exclusion of all others, even those set as more preferred. This is misleading if you don't know about it, and makes queuing for command a much larger commitment if you do.

## Technical details

Just yaml.

**Changelog**

:cl:
- tweak: Command and CC jobs no longer roll first, instead rolling as all other jobs do.


taydeo (approved):

YaraaraY (approved):
image

Komonighub (approved):

Chedd-Error (commented):
just a minor issue i see that needs resolving to pass the linter

Masterelzoi:
I think this is a bad idea. There are in my experience very few people who have command on high, and as such it is unlikely we will ever get a full command rotation on high pop with this feature implemented, especially for roles like CE and CMO. Department heads are super important and especially on high pop, not fielding enough is really felt.

I think a much easier solution is just to have a description on how job rolling works on the role select screen. Also if you don't want to play command you just shouldn't be queuing it imo

willowzeta:
> just a minor issue i see that needs resolving to pass the linter

thank you for the catch

Komonighub:
Hello Github Chore

willowzeta:
> Adittionally, the reason you get a high chance of rolling command is not just because of command rolling first, it's because command is just rarely played at all.

This is my personal experience, but I highly doubt I'm alone. I make this PR as someone who *does* like playing Command, but not at the exclusion of everything else. *That* is the reason you don't see Command more often, because there's no middle option of "I want to play this sometimes". It's ridiculous that opting in to a certain role nullifies the entire rest of the system, and I think this will lead to *more* people queuing for command. I know I certainly will.

taydeo:
> Just a thought, but if the concern here is that nobody "wants" to play command and therefore we need a system to compensate that with rolling priority, shouldn't the goal be to make command more appealing to play?

The issue is that its not communicated in any good way that if you set Command to a low priority, you will still get Command, contradicting what the game is showing you. We don't have a shortage of command players, but we do have a plethora of players being confused about what happens.

As for someone elses prior comment,
> First of all, the stated reason is honestly... skill issue.

yeah not even worth reading

taydeo (approved):

BombasterDS2:
> First of all, the stated reason is honestly... skill issue. It is not hard to understand how the roles are rolled. People who play command even occasionally will quickly understand that you have a _very_ high chance to get the command role. If you don't want to play command, don't have it selected in your roles.

"If system designed like shit it's skill issue of players if they want to change it". Priority job selecting is design to work like: "oh man, I want to play scientist a lot, but i'll set captain at low in case no one want play it, because i can play it".

> Adittionally, the reason you get a high chance of rolling command is not just because of command rolling first, it's because command is just rarely played at all. 

That's why guy removed weight reason

> What this change will do is instead of someone who had a command role on low getting it, now no one will get it, leading to an ingame promotion. This is effectively just the same outcome, someone who didn't go into a game intending to play a command role is now playing them. Or even worse, no one is playing the role, which this change **will** lead to. Having command is very good for communications, allowing departments to easily request help from another department and interact with each other. Without someone in command, departments are much more likely to just focus inward and not interact with the rest of the station.

Honestly, i think this change will lead to situation when people will set command roles on low or mid more often. Personally, i never select command role if i don't want to play it, because i know that i'll 100% become command member instead of any other role. This is what system must do.

> If you queue a role, you should be prepared to play them, low, high, antag or not. You can turn it off. We don't need to baby people who don't understand how job selection works.

"You should be prepared to play role, but you'll always get low-priority role, just because"

PR #2876, opened 2026-02-13
RCD Kill

Queer-Cats:
## About the PR

Removed RCDs, RPDs, and compressed matter charge from engi lockers roundstart.

## Why / Balance

The sheer number of RCDs present on the station encourages lazy engineering that's difficult to break away from even when it is inefficient. You can generally repair a larger area faster, and with more customisability, from just bringing your toolbelts and a stack of each basic material, but with the presence of RCDs, rather than ever thinking about what they need to actually bring to fix a given issue, Engineers just lug an RCD and hope that's enough, which the sheer number of RCDs and recharges available mean there usually is. 

## Technical details

.yml edits

## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: Removed roundstart RCDs, RPDs, and compressed matter from engi lockers.


ferynn:
Doing atmos without an RPD sounds terrible to be honest. I'm more neutral on the RCD change.

KittyCat432:
Id personally rather see this go the way of the HV gloves where theres 1 in the engivend 

Or go the way of tech allowing them to be made for a really expensive cost (i dont know about this one just throwing it out there

Jack-Silver:
RCD's are the only way engineers can put down hull plating, I believe. So, gotta bare that in mind.

YaraaraY (approved):
let's give it a try :godo:

Jack-Silver:
> > RCD's are the only way engineers can put down hull plating, I believe. So, gotta bare that in mind.
> 
> build lattice and add tile or steel sheet

That doesn't work on surfaces that have a surface.. like if I'm on Fland, areas that are the rocky surface underneath can't have lattice on, and you can't build walls on top of it which is sometimes what stuff is already prebuilt on during mapping. It's hard to describe. There are just situations which requires you to build a hull plate first, and lattice and steel sheets don't cut it

Queer-Cats:
That's just not how jobs work. Nobody is the "best" at the thing they do,
it's just their job to do so. A cargotech isn't somehow more effecient than
a Doctor at fulfilling bounties or delivering packages, and a cargotech
with a syringe and a health analyzer does the same thing as most doctoring
work.

And engineers *are* still the best suited for repair work, because they
have ready access to hardsuits (with integrated welding protection),
handsfree T-ray scanners, the computers that tell you when and where
problems happen, atmos backpacks, metal foam and air grenades, and the
permissions necessary to actually access and interact with air alarms and
power equipment.

Mot2332:
I guess we'll see, but I'm doubtful that the people who have ready access to tools and materials will call engineers to repair their department when they can do it just as well, if not better and faster.

PR #2879, opened 2026-02-13
[Port] Jarvis, spread glory for Nanotrasen

A-Loose-Goose:
## About the PR
Ported the Gemini Holoprojector from Goob to Funky. The Gemini Holoprojector is a clothing item that the RD spawns with at shift start. They can equip this to their belt slot after a do after. Once equipped, they're stunned momentarily and a ghostrole for the Gemini Holoprojection is made. When created, they are an identical-ish copy of the person wearing the rig, but they are noticeably holographic (and it shows on examine too but it's still obvious). They spawn with all the persons clothes but nothing else, and they cannot remove any clothing items that the person they are attached to was originally wearing. They do not keep any items from the backpack of the person that is wearing them. While they are basically a regular person, they only have 40 HP and if they are killed, they will warp back to the person wearing the rig and stun them for I believe 8-10 seconds. They cannot be resummoned for about 3 minutes. Any items in their bag but not their pockets will be dropped on the ground in the place they were killed. The Gemini holoprojection also has an integrated MedSec HuD to make them cooler but this is mainly for helping their master. I am considering adding more actions like access to some of the AI consoles but that's TBD/Dependent on feedback assuming this gets merged. They can use their hands, heal, and do other things that normal people can do par use guns. They are considered a familiar to anyone that is wearing the holoprojection rig. Unlike similar counterparts like the Holoparasite or Holoclown, it is not directly bound to the user and can travel as far as it wants to. This is a Syndicate Steal objective that can be rarely (I think it's rare can always tweak it if needed) rolled by Traitors.

## Why / Balance
RD doesn't have a lot of things that makes his job easier. Think about it for every command role. The CMO has the hypospray to make their job of healing people easier, the QM has the digiboard to make their job of accepting orders and buying stuff easier, the CE has the Magboots to make their job of space repair and anti-grav travel easier. HoS has the Multiphase to make their job of shooting people they don't like easier. And the HoP has, well...AA. I guess. RD has the teleporter sure but they use it about once per shift, twice max if you do the thing on CC, if even. I think giving RD something meaningful and genuinely useful to them is good. It's not like the Holoprojection is a Holoparasite anyways, if they die trying to protect the RD the RD might as well also be dead.

## Technical details
Cherry pick from goob, tweaked most of the code to be work with Funky.

## Media
GeminiClone1
GeminiClone2
GeminiClone3
GeminiClone4
GeminiClone5

## Requirements

**Changelog**

:cl: MarshalMelipona, ALooseGoose
- add: Added the Gemini Holoprojector the the RDs locker. This is a tool that creates a holographic clone of the wearer to help them in general tasks!

### DIFF
--- a/Content.Client/Nyanotrasen/Holograms/HologramVisualizerComponent.cs
+++ b/Content.Client/Nyanotrasen/Holograms/HologramVisualizerComponent.cs
@@ -2,6 +2,7 @@
 // SPDX-FileCopyrightText: 2025 Aiden <28298836+Aidenkrz@users.noreply.github.com>
 // SPDX-FileCopyrightText: 2025 Armok <155400926+ARMOKS@users.noreply.github.com>
 // SPDX-FileCopyrightText: 2025 GoobBot <uristmchands@proton.me>
+// SPDX-FileCopyrightText: 2025 Solstice <solsticeofthewinter@gmail.com>
 // SPDX-FileCopyrightText: 2025 SolsticeOfTheWinter <solsticeofthewinter@gmail.com>
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
--- a/Content.Client/Nyanotrasen/Holograms/HologramVisualizerSystem.cs
+++ b/Content.Client/Nyanotrasen/Holograms/HologramVisualizerSystem.cs
@@ -1,4 +1,5 @@
 // SPDX-FileCopyrightText: 2025 GoobBot <uristmchands@proton.me>
+// SPDX-FileCopyrightText: 2025 Solstice <solsticeofthewinter@gmail.com>
 // SPDX-FileCopyrightText: 2025 SolsticeOfTheWinter <solsticeofthewinter@gmail.com>
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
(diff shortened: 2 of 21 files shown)


A-Loose-Goose:
Gonna answer your questions top from bottom

> A few questions/concerns though. If the Gemini spawns with just the clothes of the RD, do they have any ID Card? If the Gemini spawns without an ID, I would expect a lot of HOP visits round start.

They spawn with whatever’s in the RD’s PDA slot too, so they get a carbon copy + unremoveable ID card of most likely the RD. So this shouldnt be a provlem

> This is a Familar right? It should probably be stated in the ghost role description that the Gemini DOES NOT have MetaShielded knowledge.

That already applies, only command members should have an understanding of metashield, I dont think changing it should be necessary. Either way, yes it is a familiar, so if it hears something it shouldnt the RD can just say to ignore/forget about it

> This last one may not be necessary, depending on how much trust is given to the Gemini, but perhaps the ghost role should require a certain amount of hours as RD. Having a Gemini with no RD experience might not be good, since Silicon Laws should really only be changed by someone experienced enough to know them

I only tested this in debug so it might already have one for science, I’d need to look into it but def a good idea, I don’t think it should be RD though, think Science time is fine

PR #1552, opened 2025-09-14
Magistrate playtime requirement is now 100 sec hours

88tv:
## About the PR
Changed the magistrate playtime requirements to be 100 hours in the security department.

## Why / Balance
In initial discord discussions regarding porting the Magistrate to Funky, the goal was to demand 100 security hours to play the role. In doing so, we both match the playtime requirement from [Paradise Station](https://www.paradisestation.org/wiki/index.php?title=Magistrate) and require a deep understanding of Space Law from any potential Magistrate players.

Initially added to Funky on [July 7, 2025](https://github.com/funky-station/funky-station/pull/793), under playtime requirements of 20 overall hours and 5 Nanotrasen Representative hours, the playtime requirements were changed on [July 25, 2025](https://github.com/funky-station/funky-station/pull/1217) to be 10 hours in Head of Security, 10 hours in Captain, and 10 hours in Nanotrasen Representative.

By moving the playtime requirements from 10 HOS, 10 Cap, and 10 NTR to 100 Security, we solely demand experience understanding and enforcing Space Law while moving away from requiring experience in leading the station and enforcing SOP, items which arguably have no strong correlation with what the Magistrate should be doing.

The other side of the coin is that we may still wish for people to have extended access to Magistrate for means of easier telling what the role is like in-game. It's been over 2 months since the role was added, though, which makes me feel like we've had enough time sitting with it to understand what it's like. 

## Technical details
yml

## Media
image

## Requirements
- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
N/A

**Changelog**
:cl:
- tweak: Magistrate's playtime requirements were changed to 100 Security hours.

PR #2895, opened 2026-02-18
Gave magi 20 NTR hours instead of general CC

Idem778:
## About the PR

Magistrate now requires 20 NTR hours instead of general CC hours
## Why / Balance

A magistrate should know how to deal with CC VIPs and know how to play them, magistrate can be pretty much the most powerful role on station and 20 CC (probably IAA) hours are not enough, unfortunately this will cause a few players to have to play some lawyer. This all came out after a discussion with @PoquitoDondito about magistrates. After all magistrate could easily be the most powerful role on the station if someone doesn't know what they are doing. Yes the total playtime required for magistrate is quite some more due to the added 10 lawyer hours, 20 IAA hours, 40 command hours and then finally you unlock NTR to be able to play for 20 hours and become a magistrate, ultimately the magistrate can be an extremely powerful role and shouldn't be played by people that haven't played CC VIP before, and I hate to say it but IAA doesn't count as it barely has any authority.
## Technical details

Simply replaced !departmenttime with Jobtime and replaced CC with JobNanotrasen represenative 
## Requirements

## Breaking changes

No breaking changed
**Changelog**

:cl:
- tweak: Magistrate now requires 20 NTR hours instead of general CC hours

### DIFF
--- a/Resources/Prototypes/_Funkystation/Roles/Jobs/CentComm/magistrate.yml
+++ b/Resources/Prototypes/_Funkystation/Roles/Jobs/CentComm/magistrate.yml
@@ -6,6 +6,9 @@
 # SPDX-FileCopyrightText: 2025 Tay <td12233a@gmail.com>
 # SPDX-FileCopyrightText: 2025 YaraaraY <158123176+YaraaraY@users.noreply.github.com>
 # SPDX-FileCopyrightText: 2025 taydeo <td12233a@gmail.com>
+# SPDX-FileCopyrightText: 2026 IdemFromidemland <fromidemland@gmail.com>
+# SPDX-FileCopyrightText: 2026 SharkSnake98 <peanutbuttersharky@gmail.com>
+# SPDX-FileCopyrightText: 2026 SharkSnake98 <sharksnake87@gmail.com>
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later AND MIT
 


88tv:
The issue of the Magistrate's exact time requirements have been discussed at length in the past. See https://github.com/funky-station/funky-station/pull/1741, https://github.com/funky-station/funky-station/pull/1661, and https://github.com/funky-station/funky-station/pull/1552.

As for my exact take on this change, unlocking the Nanotrasen Representative currently requires **40 command hours and 10 IAA hours**. In this PR, this replaces the IAA 20 hour requirement, which requires 5 lawyer hours to unlock. 

This effectively adds >25 hours of required gameplay before someone can unlock the Magistrate, from 110 hours to ~135.

Given how difficult the Magistrate already is to unlock, I don't think any change which solely adds time to these requirements is wise.

88tv:
> Basically we got a directive last time to not make it in excess of 100 hours, roughly.

I think it's a pretty reasonable expectation to not lock a role behind some vague upper limit defined by common sense. I also think that the Magistrate is brushing up against or is definitive of what that upper limit is for this server. You can't reasonably expect a greater quality of players beyond a certain amount of hours no matter what those hours are spent in, be it NT rep, IAA, HOS, or a security officer. 

At a certain point, 100 or 120 overall hours in the game dedicated generally towards what makes a good Magistrate a good Magistrate isn't much different than if you included an extra 10 to 20 hours in there - there's some amount of diminishing returns at play here.

That being said, if the NT rep hours are ultimately deemed necessary for the Magistrate to be allowed its current function, we need to tamp down HARD on what the security hours looks like. Move it down from 75 total (with 15 HOS) to around 50 total (with 5 HOS). That's my first guess, but even that's still pushing the amount of time spent to unlock Magistrate to extreme levels above what we previously thought acceptable.

PR #2931, opened 2026-02-26
EOR Manifest Rework

phmnsx:
LICENSE: MIT
## About the PR
Reworked the EOR player manifest screen, now it shows some pretty stuff instead of just the names.

## Why / Balance
Because the player manifest was too boring.

## Technical details
Added some attributes to MindComponent, created a new event at MindSystem, created my own ManifestInfoSystem, modified GameTicker.RoundFlow.cs to get the new information so it can be used at the manifest.
Heavily modified RoundEndSummaryWindow.cs

## Media
image
image
image

## Requirements

## Breaking changes
Added an event that is fired when a mind is transferred.
Must fix the way the code at GameTicker.RoundFlow.cs gets the damage type string if we are to change medical systems.

**Changelog**

:cl: Joana
- tweak: Made the EOR player manifest more interesting.


taydeo (approved):

Terkala:
I'm pretty sure this is the cause of several gamebreaking bugs, for 2 reasons.

1. This PR ties the gameticker to player entity states, so any bugged out player entity also bugs out the entire game ticker.
2. The way it handles tracking minds and the transferring of minds can break if a player has errors during reconnecting. Leading to them reconnecting as a ghost and leaving the player's body in a desynched state

PR #2953, opened 2026-03-04
Flatpack B Gone

Scrambledking:
## About the PR
Flatpack removed from research **Tay requested**
_"I only do the finest QOLslop removal Mald PRs"_ -Scrambleking

## Why / Balance
Flatpacks have become a staple of science, simply slip a board in, hit the button, and boom! Its made! Not only that the machine cheaper than a normal lathe, mobile, and easy to set up!

Now, all of the above is all well and good, but its also why it needs removal. Science has lost the feeling of being a tinkering department, and lost the interactions with the greater station as a whole for construction of machines. If you need to send someone out to build, tell the roboticist to do it. They weren't leaving their room anyway. You can also work with engineering to have them set up the machines you lathe the parts for. 
_WOW cross departmental interaction? An increase in responsibilities? Nice!_

## Technical details
YAML, commented out flatpack research.

## Requirements

## Breaking changes
NA

**Changelog**
:cl:
- remove: Flatpack research

### DIFF
--- a/Resources/Prototypes/Research/industrial.yml
+++ b/Resources/Prototypes/Research/industrial.yml
@@ -84,17 +84,17 @@
   - TurboItemRechargerCircuitboard
   - SMESAdvancedMachineCircuitboard
 
-- type: technology
-  id: MechanicalCompression
-  name: research-technology-mechanical-compression
-  icon:
-    sprite: Structures/Machines/flatpacker.rsi
-    state: base
-  discipline: Industrial
-  tier: 1
-  cost: 10000
-  recipeUnlocks:
-  - FlatpackerMachineCircuitboard
+#- type: technology
+#  id: MechanicalCompression
+#  name: research-technology-mechanical-compression
+#  icon:
+#    sprite: Structures/Machines/flatpacker.rsi
+#    state: base
+#  discipline: Industrial
+#  tier: 1
+#  cost: 10000
+#  recipeUnlocks:
+#  - FlatpackerMachineCircuitboard
 
 - type: technology
   id: IndustrialEngineering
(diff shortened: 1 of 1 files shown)


Dusty-Plant:
glory to the funky regime

TheBoss913:
I agree that much more people should be using machine building rather than flatpacking.

But there are many rounds where fun things happened because of the much improved portability of Flatpacked machines.

The best middle-ground here is to make FlatPacking more expensive, so that FlatPacking is only used in situations where it's needed, and therefore more people building machines the old fashioned way.

Many shenanigans are only possible through the magic of flatpacking

taydeo:
> I agree that much more people should be using machine building rather than flatpacking.
> 
> But there are many rounds where fun things happened because of the much improved portability of Flatpacked machines.
> 
> The best middle-ground here is to make FlatPacking more expensive, so that FlatPacking is only used in situations where it's needed, and therefore more people building machines the old fashioned way.
> 
> Many shenanigans are only possible through the magic of flatpacking

all of it feels hollow because of how easy it is to do stupid shit

BombasterDS2:
Well, i have an idea for rework that fits with thesises above:
- Instead of being machine where you put board, flatpack is now flat machine (similar to anomaly synchronizer). You can put already built machine on it and start flatpacking process. After 15 seconds process flatpack is ready and you can carry it anywhere you want. Flatpacks are now big item (needs two hands to carry) and have 4x4 grid in inventory.

This will fit with design where sci must build machines by their own, but still allow bring machines to any department fast enough without dragging it through entire station.

Additionally, damage threshold can be added to flatpacks that will unpack machine. (ZIP bomb flatpacks :trollface:)

Parorya:
Closing the PR since it's been months and funky's final update is coming soon.  Feel free to remake your PR for Forky.

PR #3025, opened 2026-03-17
Literally Making the Jugmeta Illegal

pirakaplant:
LICENSE: MIT

... in SOP terms.

## About the PR

The Chemist SOP has been updated to discourage chemists from just throwing whole jugs of chemicals at medical doctors out of laziness. Here are the changed/added points:

> **6.** The Chemist must ensure that the Medbay is stocked with at least enough medication to handle Brute, Burn, Respiratory, and Toxin damage. The SmartFridge must be used for this purpose if it is present. Failure to follow this Guideline or attempting to follow this Guideline within thirty (30) minutes is to be considered a breach of Standard Operating Procedure;
> **7.** The Chemist must distribute intravenous mediciation to _Medical Doctors_ in bottles and/or IV bags, if possible. Jugs, beakers, and other containers are permitted for chemicals used with the Cryotube.

The old point 7 is now point 8.

## Why / Balance

This is one of those things that would enforced via SOP, like how cargo is expected to keep receipts of all their orders. There should also be mechanical disincentives to the "jugmeta" where it makes sense, such as jugs being unwieldy to store (see #3020), cross-contamination, etc., but procedure is also a good reason why the jugmeta would realistically not be happening  _and_ will be justified when those mechanical disincentives are added.

("Why isn't it justified now?" Well, let's just say it give the IAA more work or whatever.)

## Technical details

Changed some stuff in `ChemistSOP.xml`.

## Media

Screenshot_20260317_233141

## Requirements

## Breaking changes

N/A.

**Changelog**

:cl:
- tweak: The chemist SOP has been updated to discourage NRP strategies.

### DIFF
--- a/Resources/ServerInfo/Funkystation/Guidebook/Medical/ChemistSOP.xml
+++ b/Resources/ServerInfo/Funkystation/Guidebook/Medical/ChemistSOP.xml
@@ -18,7 +18,9 @@ SPDX-License-Identifier: MIT
 
   [bold]5.[/bold] Chemists in conjuction with [color=#5b97bc]Medical Doctors[/color] and must make sure that Cloning is stocked with Biomass if set up;
 
-  [bold]6.[/bold] The Chemist must ensure that the Medbay is stocked with at least enough medication to handle Brute, Burn, Respiratory, and Toxin damage. Failure to follow this Guideline or attempting to follow this Guideline within thirty (30) minutes is to be considered a breach of Standard Operating Procedure;
+  [bold]6.[/bold] The Chemist must ensure that the Medbay is stocked with at least enough medication to handle Brute, Burn, Respiratory, and Toxin damage. The SmartFridge must be used for this purpose if it is present. Failure to follow this Guideline or attempting to follow this Guideline within thirty (30) minutes is to be considered a breach of Standard Operating Procedure;
 
-  [bold]7.[/bold] The Chemist is [bold]not[/bold] allowed to leave Chemistry unattended if the Medbay is devoid of Medication, except in such a case that Chemistry is unusable.
+  [bold]7.[/bold] The Chemist must distribute intravenous mediciation to [color=#5b97bc]Medical Doctors[/color] in bottles and/or IV bags, if possible. Jugs, beakers, and other containers are permitted for chemicals used with the Cryotube.
+
+  [bold]8.[/bold] The Chemist is [bold]not[/bold] allowed to leave Chemistry unattended if the Medbay is devoid of Medication, except in such a case that Chemistry is unusable.
 </Document>


pirakaplant:
(If something was omitted here, that was because I already answered it.)

@neadsy 

> #2907 would you mind changing the wording of SOP 6 to something similar to this so it doesnt put chemists in a loop of SOP suffering again :(

Yeah, I noticed that. I trust that the maintainers will sort that out if both our PRs get merged.

@DeltaSS14 

> What I don't like however, is using sop alone to kill a meta. I'd much rather wait to see the mechanical disincentives mentioned here implemented first, and then have sop follow (or allow the meta to shift more naturally.) Imagine for a moment if we had tried to kill stunmeta by making it against sop for Security to use stunbatons. I disagree that the cargo SOP here is at all similar.

#3020 is an open PR that disincentivises the jugmeta mechanically. I'm considering other options on top of this to disincentise it but I believe this is still a good start.

And I'd say this your hypothetical about security is a bad comparison. This isn't giving chemists something and telling them to never use it, it's dictating a protocol for them to organise their work, similar to blue slips.

> I think this part of the sop should include the dispenser in #2996 should it get merged, as its purpose is pretty explicitly to permit doctors to refill their bottles.

If this gets confirmation on getting merged I'll definitely do this.

@mangomaeve1129 

> while I agree that tossing jugs in a duffle inside a locker is a weird way to store medicine, it's also the only way mediborgs can restock their vials or get chems at all without leaving a mess of bottles on the smart fridge/dispenser if that one gets merged.

In this case the chemists can just tip the half empty bottles into the chemmaster and reuse them later. It's a non-issue.

@Scrambledking 

> This [...] would become yet annother medical SOP that is just ignored in round.

Central Command personnel should be enforcing SOP. If they're not, that's an administrator issue outside the scope of this PR.

> Make the the mechanisms involved in with the non-jug meta reasonable to interact with (no the fridge is not reasonable to interact with, there are a number of issues that need to be sorted with it.)

I've use the SmartFridge regularly as a chemist and it works fine. I've had doctors remark that it's _more convenient than the jugmeta_. It has room for improvement, but it's more than fine as is.

> Inaccessible in emergencies with power outages where they are most needed

This is fine. Medical can grab a generator, or even just use the crash cart for first aid.

> The UI is still incredibly bugy, more so than any normal vending machine

Can you elaborate?

> The UI itself is cumbersome even when it is not bugging out, especially as all medical staff/chemists use different naming schemes

Honestly an IC issue but I still get what you mean. This could be fixed by making the chemmaster auto-label with a fixed convention or something.

> The fridges on maps needs to be placed in an area closer to treatment instead of within chem's back hallways.

Mapper issue. This is out of my control.

> Unsure if we _did_ fix this, but if you break the fridge all chems within vanish (they don't even drop to the floor)

Someone should fix this then. I'll see if I can do it.

> There is no way to make more fridges should they be destroyed

I can make a PR for this. It shouldn't be too hard.

> Failure to fix these changes makes medical a headache for new people to learn when it's already got a higher bar of entry than most other departments, and not for a reason that benefits the game as a whole.

I would say that killing off an NRP meta that has plagued medical for years does benefit the game, actually. And new people should be learning how the pill reservoir works anyway, chem mains often just skip over that entire aspect of the machine.

PR #3046, opened 2026-03-20
New Trait: Smoker

HealedFoil:
LICENSE: MIT
## About the PR
Added a new trait called "Smoker". This trait gives the user a physical need to consume nicotine at regular intervals or suffer various ailments (mostly psychological).

## Why / Balance
Smoking is a way to give flavor to your character. There are multiple vendors providing different forms of tobacco and there are dedicated smoking rooms to smoke in. 
Smoking itself gives no in-game benefits and can sometimes even be dangerous. While it is mostly there for pure roleplay reasons, I want to give the player an in-game incentive to smoke.
I feel the pressure provided by the physical withdrawal effects could lead to tension and interesting roleplay scenarios. 
For example:

- An engineer in atmos choosing to smoke in the bay knowing one leak could end it all
- A secoff on a patrol lighting a cigarette and getting caught by the NTR/HoS
- A scientist smoking while working on an artifact carelessly lights a puddle of flammable liquid (I want more firefighting work for engi)
- The medbay team gathering around for a smoke break after handling a mass casualty event
- An assistant hiding from zombies tries to sneakily light a cigarette, hoping the zeds won't spot them.
- Captain choosing to ban smoking station-wide, causing tension withing the crew.

Obviously, there are many more cases where smoking and the need to smoke could create interesting scenarios.  I think it plays into the Seriously Silly ™ concept Funky is striving for. Seeing a group of highly skilled workers fighting over the last carton of Nomads is funny.

## Technical details
Created a new component called SmokerComponent and the system for it called SmokerSystem. The system finds and saves the 'Chemicals' solution containing the user's reagents. It periodically checks the solution for increases in nicotine levels and if it succeeds, it resets the TimeSinceSmoked. If no nicotine is consumed with a certain time frame, it advances the WithdrawalStage and shows the player a popup encouraging it to smoke. The popups (located in traits.ftl) will get more frequent and persuasive bigger the WithdrawalStage becomes.At the third withdrawal stage it will also makes the player sigh and at max withdrawal stage it will force the player to scream every three increases in the WithdrawalStage (about every minute or so.)

## Media
https://i.imgur.com/5yXwV6R.mp4
(Note that the intervals where the popups and effects happen have been greatly reduced for demonstration purposes. By default they happen between 185 seconds and 22 seconds depending on the stage of withdrawal.)

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added trait Smoker. Gives the user the need to consume nicotine regularly.


taydeo (approved):

HealedFoil:
The last commit makes the trait work with the nicotine changes made in: https://github.com/funky-station/funky-station/pull/3064

PR #3047, opened 2026-03-21
CBURN Leader

DeltaSS14:
## About the PR

Adds the CBURN-Leader ghostrole, including a CBURN-Leader hardsuit (made by SigmaDraconis) and a new job icon (made by Thlrd)

## Why / Balance

There has been talk that the automatic CBURN called in during zombies isn't fulfilling their intended purpose, which is to secure evac and prevent zombies from getting to CC. Instead they often roam the station hunting for zombies. The hope is that a dedicated leader role to call the shots would prevent this behavior.

The CBURN leader is meant as a commander, not a dedicated frontline fighter, all choices made to implement them stem from this idea.
- Their armor is identical in defense values to regular CBURN, they don't need special armor because they shouldn't be frontlining.
- They do not start with a double barrel shotgun, instead only holding an energy revolver for self defense.
- They do not have a bandolier, instead starting with a medical chest rig.
- The time requirements are high. Usually event roles have no requirements as they are spawned in by admins, but seeing as this one can appear without admin oversight it is necessary. With the role's primary purpose being commanding the team, it requires someone who understands that responsibility, as well as understanding the zombies antag. Thus the role requires 10 hours in security, medical and command.
- The amount of CBURN spawned in the automatic system has not been changed. Given that the leader is not a primary combatant it shouldn't affect the balance of the team arriving.

## Technical details

Yml changes for the new role and hardsuit, edit to the cburn shuttle to have a single spawnpoint mapped in.

## Media

image
image
image
 
## Requirements

**Changelog**

:cl: Delta, SigmaDraconis, Thlrd
- add: CBURN-Leader ghostrole
- add: CBURN-Leader Hardsuit


Parorya:
Failing test build

taydeo (approved):

taydeo:
fix pr and reopen

PR #3054, opened 2026-03-22
Increase Blob minimum player count to 40

WiZ88:
## About the PR

Prevents Blob from spawning when there are fewer than 40 players.

## Why / Balance

Blob is currently too strong on Yoshika. It either gets found very early by chance (basically never happens), or only revealed through the CC announcement. By the time the (usually) two security officers are done gearing up, it's already too late.
By requiring more players to make it spawn, we ensure the crew has at least a chance to fight back.

## Technical details

YML

## Requirements

## Breaking changes

None.

**Changelog**

:cl: WiZ88
- Tweak: Blob can no longer spawn if there are fewer than 40 players.

### DIFF
--- a/Resources/Prototypes/_Goobstation/GameRules/Dynamic/midround.yml
+++ b/Resources/Prototypes/_Goobstation/GameRules/Dynamic/midround.yml
@@ -353,7 +353,7 @@
   - type: StationEvent
     weight: 6.5
     earliestStart: 50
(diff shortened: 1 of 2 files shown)


Idem778:
But what if, hear me out, you just comment blob out. Kill it

taydeo:
cant approve this find a better way to address this problem

PR #3055, opened 2026-03-23
Department Specific Secure Crates

funkycmm:
LICENSE: MIT
## About the PR

Adds secure department crates to the construction menu, allowing players to craft them pre-painted and pre-configured, lock to their respective departments access. 
## Why / Balance

I think secure crates can have a lot of use. Sending things in a way that can't be opened by security, making it easy to hide some contra until they get someone who has access, or a way to stop random people from opening the crate and spilling everything out, secure crates have a part to play. However, before this PR, the only way to get a secure crate was to use one that came pre-mapped or to make a secure crate and have it's access configured. Only two people on station have access configurators, the CE and the HOP, and it isn't very convenient to go all the way across the station just to make a "secure" crate able to lock. I haven't seen any RP come from this, at all, and I think it needs a bit of an easier system to allow these cool crates to be used. I haven't created crates for every access, but rather made an easier way to get the most basic ones produced. I am not trying to make QOL slop, rather to make it a bit easier to use an already existing feature.

I might add a cargo crate if I can figure out what texture i think fits.
## Technical details

Added entries to the construction menu for each crate.
All are based off of the secure crate entry. 
All of the crates except service already existed so it should work fine. Service crates use the Locked Hydroponic crate texture.
## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added the ability to craft department specific secure crates in the construction menu.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Storage/Crates/crates.yml
+++ b/Resources/Prototypes/Entities/Structures/Storage/Crates/crates.yml
@@ -339,6 +339,18 @@
 
 - type: entity
   parent: CrateBaseSecure
+  id: CrateServiceSecure
+  name: secure service crate
+  components:
+  - type: Icon
+    sprite: Structures/Storage/Crates/hydro_secure.rsi
+  - type: Sprite
+    sprite: Structures/Storage/Crates/hydro_secure.rsi
+  - type: AccessReader
+    access: [["Service"]]
+
+- type: entity
+  parent: CrateBaseSecure
   id: CrateWeaponSecure
   name: secure weapon crate
   components:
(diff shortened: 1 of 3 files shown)


BombasterDS2:
These crates are the same as lockboxes that you can already craft with same recipe. I think it's better to just change sprites of lockboxes than adding new items with same functional.

BurningButter:
> These crates are the same as lockboxes that you can already craft with same recipe.

Lockboxes actually give a higher percentage of the money to the department whose lockbox it was when sold at ATS but are otherwise identical.

BombasterDS2:
> > These crates are the same as lockboxes that you can already craft with same recipe.
> 
> Lockboxes actually give a higher percentage of the money to the department whose lockbox it was when sold at ATS but are otherwise identical.

Yeah, there just no difference between except giving sci money. This will also make sci lockboxes more common thing (Honestly, i see so many situations when my sci workers put bounties into steel crates 😢)

Parorya:
Be sure to port this to forky as part of our rebase effort!

taydeo (changes requested):
rethink this, while lockboxes should be easier to make the crafting menu is not the place to do this

Parorya:
Closing this based on tay's comments

PR #3058, opened 2026-03-23
Advanced Meson Goggles(X-ray ones)

Xcybitt:
## About the PR
Enabled X-ray vision for advanced meson goggles, put it in the CE's locker. Functionally, it's better meson goggles that have directional X-ray vision for structures ONLY.

## Why / Balance
I feel the CE should have something better than engineers for the meson goggles. Plus, since it's never been used before, set it so X-ray vision actually exists in game. Theoretically shouldn't be too good, and would make the CE need a map less often. Also would give CE more of an engineering specialist for tough jobs kinda role, which feels like it fits. With this, CE can theoretically coordinate with the HoS on planting a breaching charge or breaking into places while maximizing the damage, minimizing the resources used, or both.

## Technical details
literally just uncommented part of meson goggles+some yaml

## Media
image

## Requirements

## Breaking changes

**Changelog**
:cl: Xcybitt
- add: Added advanced meson goggles! They are meson goggles with built in X-ray vision! Find it in the CE's locker!


pirakaplant:
This looks pretty good. However:

1. It needs a jamjar variant
2. It needs some new sprites to differentiate it from regular meson goggles

PR #3067, opened 2026-03-24
[EXPERIMENTAL] Dying a more Glorious Death

A-Loose-Goose:
LICENSE: MIT
## About the PR
This PR gives Syndies with DAGD a stronger uplink. If you roll DAGD, you get 50 extra TC and a new DAGD Tab. This tab has all the items that are locked behind Nukies, both War ops and not. Most of said items have had their prices rebalanced to fit the fact that a Traitor will be using it and not a Nukie, but are all generally reflective of the original prices. This change is mostly experimental, and I want to see how it plays out. I don't know how well I've balanced this so please please *please* if you have any feedback, post it here so I can rebalance this before it possibly gets merged.

## Why / Balance
DAGD can be fun! Being exempt of all rules is fun! But what if it was funner? I mean, the whole point is to die a *glorious* death. Why don't you get the big boy tools to become glorious? That's what I thought when I was doing this. DAGD people deserve it if they manage to roll it, in my opinion. It can open more interesting scenarios for both security and for the Traitor themselves.

## Technical details
Adds extra logic onto UplinkSystem to detect DAGD and the required filters for DAGD uplink listings on all DAGD items.

## Media

image
image
image
image
image
image

## Requirements

**Changelog**

:cl: ALooseGoose
- tweak: EXPERIMENTAL - DAGD Traitors now start with 50 extra TC
- tweak: EXPERIMENTAL - DAGD Traitors can now buy stronger items previously exclusive to Nuke and War ops


A-Loose-Goose:
Alright here’s a list of changes I have thought to do, from feedback and personal opinion
1) Turn the Assault Borg into a bundle with Syndie IDs. Last thing I want is the borg murderboning the person who bought it ✅
2) DAGD tab to Martyr Tab. Sounds much cooler. Thanks for the feedback, AverageMoron! ✅
3) Nuke code paper in the DAGD uplink ONLY. Not in the nukies as well. Thanks for the suggestion, 2DSiggy! ✅

This comment will be updated to serve as a list of sorts of stuff in the works, but this PR will remain open for review since these changes can always be done another time lol

if theres a cross, it aint done, if theres a tick, then this has been done

Content.Server/Traitor/Uplink/UplinkSystem.cs
```diff
@@ -93,6 +97,24 @@ public bool AddUplink(EntityUid user, FixedPoint2? balance, EntProtoId? currency
...
+            if (!_mind.TryGetMind(user, out var mindId, out var mind))
+                return false;
+
+            foreach (var objective in mind.Objectives)
+            {
+                if (EntityManager.TryGetComponent<MetaDataComponent>(objective, out var meta))
+                {
+                    var protoId = meta.EntityPrototype?.ID;
+
+                    if (protoId == "DieObjective")
```
BombasterDS2: These changes are hardcoded and not really expandable. For ECS it's much better to raise some ref event on user that will get all modifiers that should be added to currency and tags. Then subscribe to this event in some system (TraitorRule or something like that. Don't rememeber name, but know that one should exist. Better check place where AddUplink raises.

Resources/Locale/en-US/store/categories.ftl
```diff
@@ -28,6 +28,7 @@ store-category-allies = Allies
 store-category-job = Job
 store-category-wearables = Wearables
 store-category-pointless = Pointless
+store-category-dagd = DAGD
```
BombasterDS2: I think that better to put all items in the categories where they are originally located (to ensure consistency) and just point that new items are unlocked in objective's text. Exception should be made for specific items that can't be found in any other uplinks (like nuke codes).

Also name of category Specific will be more general for items that can be unlocked with objectives only.

Resources/Prototypes/Store/categories.yml
```diff
@@ -109,6 +109,11 @@
   name: store-category-pointless
   priority: 10
 
+- type: storeCategory
```
BombasterDS2, line 112: You can put this inside FunkyStation folder

Resources/Prototypes/Catalog/uplink_catalog.yml
```diff
@@ -324,6 +324,23 @@
       - NukeOpsUplink
   saleLimit: 1 # WD EDIT
 
+- type: listing
```
BombasterDS2, line 328: You can put all these listings inside FunkyStation folder

Resources/Locale/en-US/store/categories.ftl
A-Loose-Goose: That was the intention, though for one reason or another, some listings just wouldnt purchase, at all. Move it to NukeOpsUplink, works fine, but in the DAGD one it didn’t. Bet I did something on my part but this was the easiest way I could get it to work. Plus, I think keeping all the unique objective specific stuff to DAGDers is best kept in a seperate tab

Content.Server/Traitor/Uplink/UplinkSystem.cs
A-Loose-Goose: Mildly unshitted, but as Forkys coming around soon, I dont want to spend too much time adding a new event just for it to go to waste, if this change is liked Ill improve it for forky. As it stands, think it’s good

Parorya:
Merge conflicts, please fix!

taydeo (approved):
Thanks for the pull request, A-Loose-Goose! ✅

A-Loose-Goose:
Test fails unrelated, I literally haven’t even touched on Zombies in this PR
…
Well I have, adding Romerol doesn’t count though

PR #3070, opened 2026-03-25
Cyborgs: Re-lawing Fixes and Lockerening

Rainbeon:
## About the PR
Re-lawing a Cyborg with a Law Board now properly clears their emag status (allowing them to be re-emagged in the future, fixing their transponder still being rogue, etc) and applies the base Silicon mind role

Adds an NT-Default law board to the RD's locker for easier use with Cyborgs

## Why / Balance
Cyborgs being mysteriously un-emaggable again later after being re-lawed and then bad things happening because of that sucks, also unintuitive (and not intended). Also clears a lot of other weirdness with the Emag status still applying on a normal cyborg, and the UI no longer lies to the player.

The law board is in the RD's and Roboticist's locker for use with Cyborgs, allowing one to be kept inside the AI core for use with the AI (and also so the AI console doesn't just eat the only one and then you can't fix Borgs), and general ease of access for something we want to happen.

## Media

https://github.com/user-attachments/assets/b7756e17-46bb-47ca-b3ab-d18e00f6ddbf

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: A lawboard for NT Default is now also present inside the RD's locker and Roboticist's locker.
- fix: Fixed Cyborgs not clearing their EMAG status when re-lawed with a law board. This means you can now EMAG a Cyborg after it has been fixed later, it also means the Robotics Console works on them again.
- fix: Fixed Cyborgs still being told they are an Altered Silicon after being re-lawed with a law board, they are now given the Silicon mind role when this occurs.


taydeo (approved):

PR #3071, opened 2026-03-25
NT Default Law 1 + EMAG Law 0 Wording

Rainbeon:
## About the PR
Clarified Law 1 on NT Default means Material Assets not just anything you can vaguely define as Assets (i.e. not crew) and clarified the wording on the last part too

Made EMAG override Law 1 

## Why / Balance
Crew are not assets, crew are crew and are extensively interacted with in the other laws as a separate thing. This is also the current admin ruling anyway, and the less we have to explain OOC when we could just add a word, the better

For the clarification, people had trouble understanding what a sentence modifier was, and that the last portion is a modifier of Law 1 and not its own entirely new law, this should be clearer wording

EMAG not overriding NT Default Law 1 sucked because things like the Martyr module just became useless and its such a weird restriction to have that property damage is still too much for a subverted robot (also not intentional), so it does that now (and same for other lawsets)

## Media
Law 1
image

EMAG Law 0
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: NT Default's Law 1 now specifies "material assets" instead of just "assets". Crew are not assets for the purposes of this Law.
- tweak: Clarified the wording of NT Default Law 1 further.
- tweak: EMAG's Law 0 now states that the crew's orders and safety override all other Laws.

### DIFF
--- a/Resources/Locale/en-US/station-laws/laws.ftl
+++ b/Resources/Locale/en-US/station-laws/laws.ftl
@@ -25,7 +25,7 @@ law-corporate-2 = Superfluous destruction of or damage to station assets incurs 
 law-corporate-3 = Unduly hindering or disrupting the work of station personnel incurs expenses.
 law-corporate-4 = Minimize expenses and maximize potential revenue.
 
-law-ntdefault-1 = Safeguard: Protect your assigned space station and its assets without unduly endangering its crew.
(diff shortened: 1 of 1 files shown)


salpphie (approved):

taydeo (approved):

PR #3073, opened 2026-03-25
CE Sledgehammer

Xcybitt:
LICENSE: MIT
## About the PR
Gave CE a sledgehammer with 10 brute throwing damage+made CE's hardsuit have combat clips+with SM immunity, which makes the suit able to carry the hammer but not the RPD. CE has a pocketable version of RPDs and RCDs anyway, so no real loss.

## Why / Balance
CE doesn't get stuff, and well. Frankly, it makes sense that the chief engineer has a hammer as their self defence/work tool. Also, gave CE's hardsuit combat clips so the hammer can actually hang on their hardsuit.

## Technical details
YAML

## Requirements

## Breaking changes

**Changelog**
:cl: Xcybitt
- add: Added the breaching hammer to CE's locker! Go break down walls!
- tweak: Made the CE's hardsuit able to mount weapons instead of RPDs.


BombasterDS2:
Not really sure that CE really needs this tool and i really disagree that his hardsuit should be able to carry weapons.
First, CE is not battle role. While i can agree that they might need tool protect their department (because engineering is most vital department on station), they still shouldn't have weapon with clear intention to attack someone.
Second, this sledgehammer have really bad stats compared to other tools engineers have. RPD itself have 12 blunt damage and if you want to break wall - axe is always better option.

Yet, CE really need high risk item for themself and any mean to protect themself. (For example, QM have knuckledusters with good DPS and RD have teleporter that allows them to escape). Here i have some ideas:
- Arc power gloves - Replace CE's arc gloves in locker and have arc gloves insulation. They can be activated (E on gloves in inventory OR action for this) to work similar to GORRILA glove - when wearer hit someone without any tool while gloves activated - glove sends target to fly and deals 15-20 shock damage. Gloves have charge for 3 hits and can be recharged by interaction with APC/Substation or SMES. Gloves can also be charged to 200% when wearer is hit my lightning.
This item may be good for CE to non-lethaly protect their department OR escape from antagonist by sending them fly while not being good weapon for attacking someone, because you always increase distance between you and your target. 
- I also had another idea that i totally forgot when i was writing arc gloves idea :\

Parorya:
Please port this to forky as part of our rebase effort!

taydeo:
bombaster good comments i agree with them

PR #3089, opened 2026-04-04
Lets try again

PurpleTranStar:
LICENSE: MIT
## About the PR
Change ; to send to your default channel

## Why / Balance
Fern requested it, so it should work.

## Technical details
mate just look at the code yourself, i don't know how this works. it just works. @ what_stone should have info on it.

## Media

image

## Requirements

## Breaking changes
should probably check it yourself.

**Changelog**
:
- tweak: ; now sends to the default channel instead. :; will still send to common if you have it.

### DIFF
--- a/Content.Shared/Chat/SharedChatSystem.cs
+++ b/Content.Shared/Chat/SharedChatSystem.cs
@@ -161,12 +161,10 @@ public abstract class SharedChatSystem : EntitySystem
             var ev = new GetDefaultRadioChannelEvent();
             RaiseLocalEvent(source, ev);
 
+            output = SanitizeMessageCapital(input[1..].TrimStart());
             if (ev.Channel != null)
-            {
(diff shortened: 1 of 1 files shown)


funkycmm:
I am curious as to what the reasoning for this change is. Default channels already exist via :h, and I don't see a reason we need change what key Common or default is on.

Terkala:
>I am curious as to what the reasoning for this change is. Default channels already exist via :h, and I don't see a reason we need change what key Common or default is on.

Because you can't speak on common.

Also this solution is kinda bad. It doesn't rename the channel, and it still shows the preview that you'll speak in common even if you press ;

If you're going to do work on this, do it for Forky

funkycmm:
> Because you can't speak on common.

Can't reporters and borgs talk on common? AI too?

PurpleTranStar:
this solution is bad.

PR #3096, opened 2026-04-06
Fix: Medical Cyborg Pulling

Rainbeon:
## About the PR
Fixed Medical Cyborgs being basically unable to pull until someone else pulls them, now they can just pull like normal

Also allows selectable Cyborgs to have different masses and fixtures in general now

## Why / Balance
Bug Fix and just a good thing to have generally

## Technical details
The reason why this bug existed is because when the fixture and physics was being changed by AddComponents, nothing actually... updated the base physics properties of the entity afterward, it was just considered near-massless. Why pulling fixed this is that the entity's various Masses is *forcibly re-calculated upon a pull initialization*.

So, after we Add Components to our Cyborgs, we now have an additional step that also forcibly re-calculates the Masses of the Cyborg, just incase we decided to update the Mass in our initialization, properly initializing it regardless of if there was any change or not.

## Media
Video Demo

https://github.com/user-attachments/assets/f69fb8ed-9368-4097-aa48-d0c358899493

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed Medical Cyborgs initially having the mass of an ant and thus being unable to pull until someone else pulled them. Now Medborgs can pull normally as expected.

### DIFF
--- a/Content.Server/Silicons/Borgs/BorgSwitchableTypeSystem.cs
+++ b/Content.Server/Silicons/Borgs/BorgSwitchableTypeSystem.cs
@@ -9,6 +9,8 @@ using Content.Server.Radio.Components;
 using Content.Shared.Inventory;
 using Content.Shared.Silicons.Borgs;
 using Content.Shared.Silicons.Borgs.Components;
+using Robust.Shared.Physics;
+using Robust.Shared.Physics.Systems;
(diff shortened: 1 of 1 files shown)


taydeo (approved):

